PR #3, opened 2015-12-29
Determine product automatically for selecting SCAP guide

jcpunk:
Using anaconda's knowledge of what product and version this is we can dynamically select a more intelligent default SCAP guide.  This patch eliminates the need to customize the search location for RHEL (and other rebuilds).


vpodzime:
Very nice, thanks! I checked with the SSG guys that the Fedora's SSG shouldn't get version in the file names any time soon so this should work just fine.

PR #4, opened 2015-12-29
Allow using a file:// url for any SCAP guides pushed on the media

jcpunk:
This allows for a bit more flexible resource locations and the ability to put alternate guides on the media.

### DIFF
--- a/org_fedora_oscap/data_fetch.py
+++ b/org_fedora_oscap/data_fetch.py
@@ -17,12 +17,18 @@ __all__ = ["fetch_data", "can_fetch_from"]
 # prefixes of the URLs that need network connection
 NET_URL_PREFIXES = ("http", "https", "ftp")
 
+# prefixes of the URLs that may not need network connection
+LOCAL_URL_PREFIXES = ("file",)
+
 # TODO: needs improvements
 HTTP_URL_RE_STR = r"(https?)://(.*)"
 HTTP_URL_RE = re.compile(HTTP_URL_RE_STR)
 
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/data_fetch.py
```diff
@@ -90,8 +96,7 @@ def fetch_data(url, out_file, ca_certs=None):
     out_dir = os.path.dirname(out_file)
     utils.ensure_dir_exists(out_dir)
 
-    if url.startswith("http://") or url.startswith("https://") \
-       or url.startswith("ftp://"):
+    if can_fetch_from(url):
         _fetch_http_ftp_data(url, out_file, ca_certs)
```
vpodzime: Probably a time for a better name for this function. Could you please update the pull request with such a change? Maybe something like ``_curl_fetch()``?

jcpunk:
I can rename that function, but it comes out of code already within the project (data_fetch line 55).  I just want to be sure that is noted before taking a public function and renaming it.

vpodzime:
> I can rename that function, but it comes out of code already within the project (data_fetch line 55). I just want to be sure that is noted before taking a public function and renaming it.

I'm not sure we are speaking about the same thing here. I meant to rename the private ``_fetch_http_ftp_data()`` function to ``_curl_fetch()``.

jcpunk:
Ahh, got it.

New data pushed

vpodzime:
Looks good to me now, thanks!

PR #6, opened 2016-05-17
Use the System hub category provided by Anaconda

M4rtinK:
Anaconda now provides the Security category, including correctly set
sortOrder. So use it instead and remove the local copy.

Related: rhbz#1269211


M4rtinK:
The related Anaconda PR: https://github.com/rhinstaller/anaconda/pull/632

bcl:
That's certainly an improvement. I also think it would look just fine as part of the SYSTEM section. I don't really think we need a whole new section for just 1 addon, especially since the title of it is 'Security Policy'

vpodzime:
Looks good to me.

vpodzime:
> That's certainly an improvement. I also think it would look just fine as part of the SYSTEM section. I don't really think we need a whole new section for just 1 addon, especially since the title of it is 'Security Policy'

I would have no problem with this either. There seems to be no other initiatives that could lead to more security-related addons.

M4rtinK:
I've updated the pull request to use the System category. :)

vpodzime:
> I've updated the pull request to use the System category. :)

Doesn't really look like you did. Something must have gone wrong.

M4rtinK:
>Doesn't really look like you did. Something must have gone wrong.
Indeed! Let's try again. :)

M4rtinK:
Looks correct to me now.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -41,6 +40,7 @@
 from pyanaconda.ui.communication import hubQ
 from pyanaconda.ui.gui.utils import gtk_action_wait, really_hide, really_show
 from pyanaconda.ui.gui.utils import set_treeview_selection, fire_gtk_action, GtkActionList
+pyanaconda.ui.categories.system import SystemCategory
```
vpodzime: ``from`` missing here?

vpodzime:
> Looks correct to me now.

Not really. ;)

M4rtinK:
> Not really. ;)
Bah! Third time is the charm I guess. :)

vpodzime:
Looks good to me.

PR #9, opened 2016-09-12
Log password policy changes

M4rtinK:
Log changes to password policy done by the OSCAP addon
and also log content of the password policy list.

Related: rhbz#1365032


mpreisler:
ACK. Thanks for the contribution!

Issue #1, opened 2015-08-27
Oscap Anaconda Addon: Don't exit with traceback when provided XCCDF / DataStream benchmark / tailoring file contains non-existing rule (dangling  selector)


iankko:
Oscap Anaconda Addon currently exits with traceback when some profile in XCCDF / DS benchmark references / calls some XCCDF rule, which is not defined in that XCCDF / DS (it's an dangling selector).

While this issue should be primarily fixed in the scanner && the provided SCAP content (the scanner should be able to identify the dangling selector and refuse to proceed. The content should reference just existing rules), there exists scenario when this issue won't be noticeable in the content. Suppose the following use case:
- the content provides set of profiles, each of them containing only rules defined in the benchmark,
- some profile in the content also contains non-existing XCCDF rule commented out (it's commented out as a signal to implement it in the future),
- the content user notices this rule, performs benchmark tailoring && within this tailoring marks also this rule as selected (uncomments it),
- the result being OAA will end up with traceback trying to evaluate / remediate the system according to this tailoring file (while the original SCAP content will still work properly, since that rule isn't used. And will be officially commented out only in moment it's implemented).

Therefore OAA should as preventive measure avoid this "unexpected / hard to reproducible " tracebacks, just because some concrete tailoring use case selected a rule non-existing (not implemented) in the benchmark yet.

ybznek:
Crashing was already fixed by https://github.com/OpenSCAP/oscap-anaconda-addon/commit/fb9c485418e59cfead3ab28d8ff8cfb7a19bba48

My change is that it OAA parse oscap output a print dangling selectors error to LOG.

Is it ok, or we require a better solution?

matejak:
The implementation based on 1.2 onwards doesn't use Python bindings, so all errors take form of scanner complaining in its output/return code, and I don't think that we get tracebacks any more.

PR #10, opened 2016-10-17
Log OpenSCAP errors

ybznek:
See https://github.com/OpenSCAP/oscap-anaconda-addon/issues/1
image

How branching work in this repo? This PR could be to master, but I am not able to test it in master branch.

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -30,6 +30,10 @@ import subprocess
 import zipfile
 import tarfile
 import cpioarchive
+import re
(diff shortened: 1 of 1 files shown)


mpreisler:
> How branching work in this repo? This PR could be to master, but I am not able to test it in master branch.

rhel7 is the maintenance branch. I think we should adopt the same convention we use everywhere else and rename the branch to `maint-0.7`. What do you think?

ybznek:
@mpreisler I agree.

PR #18, opened 2017-02-23
Use the init_done signal instead of wait_all() (#1380224)

M4rtinK:
Using the wait_all() function provided by the Anaconda thread manager
singleton can result in unexpected behavior, as there migh still be
threads unrelated to spoke initialization still running when the
function is called. This might result in unnecessary delay before the
OSCAP addon spoke becomes ready and installation can continue.

So use the init_done signal, which is the proper way to wait for
Anaconda spokes to finish initialization.

Also bump the Anaconda version to one that provides the init_done
signal.

Resolves: rhbz#1380224

### DIFF
--- a/org_fedora_oscap/gui/spokes/oscap.py
+++ b/org_fedora_oscap/gui/spokes/oscap.py
@@ -214,6 +214,14 @@ class OSCAPSpoke(NormalSpoke):
 
         self._error = None
 
+        # wait for all Anaconda spokes to initialiuze
+        self._anaconda_spokes_initialized = threading.Event()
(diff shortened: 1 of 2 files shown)


M4rtinK:
PR adding the init_done signal to Anaconda:
https://github.com/rhinstaller/anaconda/pull/968

mpreisler:
ACK

Thanks for this @M4rtinK ! I need to do more testing but so far it seems to have solved https://bugzilla.redhat.com/show_bug.cgi?id=1379479

M4rtinK:
Nice! :)

PR #21, opened 2017-06-13
Parse HTML tags from SCAP content

yuumasato:
Some Profile descriptions contains HTML tags to structure and format the
content and oscap-anaconda-addon has problems interpreting them.

This patch removes all HTML tags, while also trying to maintain
readability of content.

### DIFF
--- a/org_fedora_oscap/content_handling.py
+++ b/org_fedora_oscap/content_handling.py
@@ -29,6 +29,10 @@ import os.path
 from collections import namedtuple, OrderedDict
 from openscap_api import OSCAP
 from pyanaconda.iutil import execReadlines
+try:
+    from html.parser import HTMLParser
+except ImportError:
+    from HTMLParser import HTMLParser
 
 class ContentHandlingError(Exception):
     """Exception class for errors related to SCAP content handling."""
@@ -49,6 +53,47 @@ class ContentCheckError(ContentHandlingError):
     """Exception class for errors related to content (integrity,...) checking."""
 
     pass
+
+class ParseHTMLContent(HTMLParser):
+    """Parser class for HTML tags within content"""
+
+    def __init__(self):
+        HTMLParser.__init__(self)
+        self.content = ""
+
+    def handle_starttag(self, tag, attrs):
+        if tag == "html:ul":
+            self.content += "\n"
+        elif tag == "html:li":
+            self.content += "\n"
+        elif tag == "html:br":
+            self.content += "\n"
+
+    def handle_endtag(self, tag):
+        if tag == "html:ul":
+            self.content += "\n"
+        elif tag == "html:li":
+            self.content += "\n"
+
+    def handle_data(self, data):
+        self.content += data.strip()
+
+    def get_content(self):
+        return self.content
+
+def parse_HTML_from_content(content):
+    """This is a very simple HTML to text parser.
+
+    HTML tags will be removed while trying to maintain readability
+    of content.
+
+    :param content: content whose HTML tags will be parsed
+    :return: content without HTML tags
+    """
+
+    parser = ParseHTMLContent()
+    parser.feed(content)
+    return parser.get_content()
 
 # namedtuple class (not a constant, pylint!) for info about a XCCDF profile
 # pylint: disable-msg=C0103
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/content_handling.py
```diff
@@ -50,6 +51,47 @@ class ContentCheckError(ContentHandlingError):
...
+class ParseHTMLContent(HTMLParser):
+    """Parser class for HTML tags within content"""
+
+    def __init__(self):
+        HTMLParser.__init__(self)
+        self.content = ""
+
+    def handle_starttag(self, tag, attrs):
+        if tag == "html:ul":
+            self.content = self.content + "\n"
```
mpreisler: use += for better readability

org_fedora_oscap/content_handling.py
```diff
@@ -50,6 +51,47 @@ class ContentCheckError(ContentHandlingError):
...
+        elif tag == "html:li":
+            self.content = self.content + "\n"
+        elif tag == "html:br":
+            self.content = self.content + "\n"
```
mpreisler: I am wondering how the parser works, it adds two newlines per li and ul and only one newline for br. Can you elaborate on these choices?

org_fedora_oscap/content_handling.py
```diff
@@ -29,6 +29,7 @@
 from collections import namedtuple, OrderedDict
 from openscap_api import OSCAP
 from pyanaconda.iutil import execReadlines
+from HTMLParser import HTMLParser
```
mpreisler: Could you please prepare for python3 here? Should be a very easy change. Try to import the new name, if it fails import the old name. It's a fairly standard pattern.

See https://docs.python.org/2/library/htmlparser.html

org_fedora_oscap/content_handling.py
```diff
@@ -308,6 +350,7 @@ def get_profiles(self, data_stream_id, checklist_id):
             id_ = OSCAP.xccdf_profile_get_id(profile)
             title = oscap_text_itr_get_text(OSCAP.xccdf_profile_get_title(profile))
             desc = oscap_text_itr_get_text(OSCAP.xccdf_profile_get_description(profile))
+            desc = parse_HTML_from_content(desc)
```
mpreisler: minor nitpick: just directly call it instead of saving it in desc and then overwriting desc

org_fedora_oscap/content_handling.py
yuumasato: Parser goes through the content fed, and acts on each start and end of a tag, and also on content within these tags.

The added new lines are to be on the safer side of readability of content, because `xmllint` messes with human readability in a incomprehensible way to me. Take for example description of `rhev-upstream` Profile in built datastream:

```
        <ns10:description override="true" xml:lang="en-US">This is a *draft* profile for STIG. This profile is being developed under the DoD consensus model to become a STIG in coordination with DISA FSO.<html:br/><html:br/>

<html:p><html:strong>Where is the RHV-H STIG?</html:strong><html:ul><html:li>Question: May I deploy a product if no STIG exists?<html:br/>
            Answer: Yes, based on mission need and with DAA approval.
        </html:li><html:li>Question: What do I use if there is no STIG?<html:br/>
            Answer: DISA FSO developed Security Requirement Guides (SRGs) to address technology areas. In the absence of a STIG, an SRG can be used to determine compliance with DoD policies. If there is no applicable SRG or STIG, industry or vendor recommended practices may be used. Examples include Center for Internet Security Benchmarks, Payment Card Industry requirements or the vendor's own security documentation.
        </html:li></html:ul><html:small>Source: http://iase.disa.mil/stigs/Pages/faqs.aspx#STIG</html:small></html:p>
</ns10:description>
```
This was nicely formatted in the source: https://github.com/OpenSCAP/scap-security-guide/blob/master/RHEL/7/input/profiles/stig-rhvh-upstream.xml

Without double newlines, the description output is like below, the Source reference shows it self attached to last list item:
```
This is a *draft* profile for STIG. This profile is being developed under the DoD consensus model to become a STIG in coordination with DISA FSO.

Where is the RHV-H STIG?

Question: May I deploy a product if no STIG exists?
Answer: Yes, based on mission need and with DAA approval.
Question: What do I use if there is no STIG?
Answer: DISA FSO developed Security Requirement Guides (SRGs) to address technology areas. In the absence of a STIG, an SRG can be used to determine compliance with DoD policies. If there is no applicable SRG or STIG, industry or vendor recommended practices may be used. Examples include Center for Internet Security Benchmarks, Payment Card Industry requirements or the vendor's own security documentation.Source: http://iase.disa.mil/stigs/Pages/faqs.aspx#STIG
```

mpreisler:
ACK

PR #26, opened 2017-08-25
Add Developer Guide

redhatrises:
This adds the beginnings of a hopefully-to-be-expanded Developer Guide. This moves the wiki page into the repo as most people don't check the wiki pages by default.


mpreisler:
So.. This is essentially moving the wiki page into the repo, right?

redhatrises:
Correct.

mpreisler:
Crap, the link was wrong. I will fix it separately.

PR #12, opened 2016-10-21
These patches made it into master, but not rhel7-branch.

jcpunk:
Since most of the work seems to be happening in rhel7-branch, I didn't want these to get lost over time.

PR #27, opened 2017-08-25
Allow using a file:// url for any SCAP guides pushed on the media or via an updates.img

redhatrises:


redhatrises:
Part of PR #12

PR #28, opened 2017-08-25
Dynamically handle product and product version for both RHEL and Fedora systems

redhatrises:
This combines the rhel7-branch and master branch way of dynamically handling product and product version in anaconda. The goal with this PR is to no longer have separate code for fedora and rhel products.

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -19,8 +19,8 @@
 #
 
 """
-Module with various classes and functions needed by the OSCAP addon that are not
-specific to any installation mode (tui, gui, ks).
+Module with various classes and functions needed by the OSCAP addon that are
+not specific to any installation mode (tui, gui, ks).
 
 """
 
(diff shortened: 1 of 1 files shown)


redhatrises:
Updated dynamic code from PR #12.

org_fedora_oscap/common.py
```diff
@@ -54,6 +55,13 @@
 
 SSG_DIR = "/usr/share/xml/scap/ssg/content/"
 SSG_XCCDF = "ssg-rhel7-xccdf.xml"
+if product.productName.lower() != 'anaconda':
+    if product.productName.lower() == 'fedora':
+        SSG_XCCDF = "ssg-fedora-xccdf.xml"
+    elif product.productName.lower() == 'rhel':
+        SSG_XCCDF = "ssg-rhel%s-xccdf.xml" %(product.productVersion)
```
mpreisler: This is a useless branch, it does the same as the branch below it

redhatrises: @mpreisler should we just do this:

```
if product.productName.lower() != 'anaconda':
    if product.productName.lower() == 'fedora':
        SSG_XCCDF = "ssg-fedora-xccdf.xml"
    else:
        SSG_XCCDF = "ssg-%s%s-xccdf.xml" %(product.productName.lower(), product.productVersion)
```

mpreisler: Yeah and please make it pep8 compliant.

redhatrises:
@mpreisler made suggested fixes.

mpreisler:
ACK

PR #30, opened 2017-08-29
Add a script to build an update image

jan-cerny:
This is a copy paste from the wiki.
https://github.com/OpenSCAP/oscap-anaconda-addon/wiki/How-to-test-oscap-anaconda-addon
You won't need to copy paste the code from wiki anymore
if we have this code in the repository.


redhatrises:
Ack.

PR #35, opened 2017-08-31
Fix some failing tests and pep8 complaints

redhatrises:
- Add '--reports' to expected args in tests
- Strip whitespace in str_test
- Fix various pep8 complaints

### DIFF
--- a/tests/ks_oscap_test.py
+++ b/tests/ks_oscap_test.py
@@ -73,7 +73,8 @@ class ParsingTest(unittest.TestCase):
         self.oscap_data2 = OSCAPdata("org_fedora_oscap")
         str_ret = str(self.oscap_data)
         for line in str_ret.splitlines()[1:-1]:
-            self.oscap_data2.handle_line(line)
+            if "%end" not in line:
+                self.oscap_data2.handle_line(line)
 
         str_ret2 = str(self.oscap_data)
         self.assertEqual(str_ret, str_ret2)
--- a/tests/ks_oscap_test.py
+++ b/tests/ks_oscap_test.py
@@ -55,7 +55,7 @@ class ParsingTest(unittest.TestCase):
                                           self.oscap_data.tailoring_path))
 
     def str_test(self):
-        str_ret = str(self.oscap_data).strip()
+        str_ret = str(self.oscap_data)
         self.assertEqual(str_ret,
                          "%addon org_fedora_oscap\n"
                          "    content-type = datastream\n"
@@ -66,7 +66,7 @@ class ParsingTest(unittest.TestCase):
                          "    cpe-path = /usr/share/oscap/cpe.xml\n"
                          "    tailoring-path = /usr/share/oscap/tailoring.xml\n"
                          "    profile = Web Server\n"
-                         "%end"
+                         "%end\n\n"
                          )
 
     def str_parse_test(self):
--- a/tests/ks_oscap_test.py
+++ b/tests/ks_oscap_test.py
@@ -6,6 +6,7 @@ from pykickstart.errors import KickstartValueError
 from org_fedora_oscap.ks.oscap import OSCAPdata
 from org_fedora_oscap import common
 
+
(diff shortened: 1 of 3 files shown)


redhatrises:
@openscap-ci test this please

redhatrises:
Bump for review.

tests/rule_handling_test.py
```diff
@@ -366,24 +369,20 @@ def passwd_minlen_short_passwd_test(self):
...
 
-        # warning should mention that something happened to the old password
-        self.assertIn("was", messages[0].text)
+        # warning should mention that something is wrong with the old password
+        self.assertIn("is", messages[0].text)
 
-        # doing changes --> password should be cleared
-        self.assertEqual(self.ksdata_mock.rootpw.password, "")
+        # doing changes --> password should not be cleared
+        self.assertEqual(self.ksdata_mock.rootpw.password, "aaaa")
```
dahaic: I don't know - is this correct behaviour? Root PW which does not follow our requirements should be cleared, so that user has to input a new one, right?

tests/rule_handling_test.py
```diff
@@ -585,56 +585,28 @@ def revert_mount_options_test(self):
         self.assertEqual(self.storage_mock.mountpoints["/tmp"].format.options,
                          "defaults")
 
-    def revert_password_changes_test(self):
+    def revert_password_policy_changes_test(self):
```
dahaic: Same as above - there are tests removed from this method, and I am not really sure it is good approach to remove them when they fail. We should probably check this behaviour in separate PR, if the problem is in test or in the code itself.

tests/rule_handling_test.py
redhatrises: @dahaic this is no longer the case. The main code no longer clears root password because it breaks the install. See https://bugzilla.redhat.com/show_bug.cgi?id=1263254. This test needs to be changed to reflect that root password is no longer cleared.

tests/rule_handling_test.py
redhatrises: @dahaic see above. This test needs to be changed to reflect that root password is no longer cleared.

org_fedora_oscap/ks/oscap.py
```diff
@@ -230,8 +232,9 @@ def handle_line(self, line):
         try:
             actions[pre](post)
         except KeyError:
-            msg = "Unknown item '%s' for %s addon" % (line, self.name)
-            raise KickstartParseError(msg)
+            if "%end" not in line:
```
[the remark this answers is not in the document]
redhatrises: @dahaic I will look into this further

tests/ks_oscap_test.py
```diff
@@ -54,7 +55,7 @@ def properties_test(self):
                                           self.oscap_data.tailoring_path))
 
     def str_test(self):
-        str_ret = str(self.oscap_data)
+        str_ret = str(self.oscap_data).strip()
```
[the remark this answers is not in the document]
redhatrises: @dahaic I will look into this further

redhatrises:
@dahaic thanks for helping with review!

redhatrises:
@dahaic ready for your review. See comments about passwords no longer being cleared which means some of the functionality of the tests have to change.

redhatrises:
@dahaic BTW. This is why I want Jenkins tests for rpm builds as the rpm has failed to build which has kept the package at FC22.

tests/rule_handling_test.py
dahaic: @redhatrises ah, makes sense. Thanks for finding the reference :)

tests/rule_handling_test.py
dahaic: ack

dahaic:
@redhatrises agreed, thank you for noticing and fixing this mess!

redhatrises:
@dahaic no problem. Thanks for the review! Let's get that Jenkins oscap tests up soon!

PR #45, opened 2017-09-08
[master] Update to python3

redhatrises:
- Update code and tests to support python3

### DIFF
--- a/org_fedora_oscap/content_handling.py
+++ b/org_fedora_oscap/content_handling.py
@@ -272,7 +272,7 @@ class DataStreamHandler(object):
 
         """
 
-        return self._items.keys()
+        return list(self._items.keys())
 
     def get_data_streams_checklists(self):
         """
--- a/org_fedora_oscap/content_handling.py
+++ b/org_fedora_oscap/content_handling.py
@@ -19,8 +19,8 @@
 #
 
 """
-Module with various classes for SCAP content processing and retrieving data from
-it.
+Module with various classes for SCAP content processing and retrieving data
+from it.
 
 """
 
@@ -54,7 +54,9 @@ class BenchmarkHandlingError(ContentHandlingError):
 
 
 class ContentCheckError(ContentHandlingError):
-    """Exception class for errors related to content (integrity,...) checking."""
+    """Exception class for errors related to content (integrity,...) checking.
+    
+    """
 
     pass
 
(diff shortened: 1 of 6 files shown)


redhatrises:
Ready for review.

@dahaic I don't think that master tests should run on RHEL7 with RHEL7 running python2.

org_fedora_oscap/content_handling.py
```diff
@@ -32,7 +32,7 @@
 try:
     from html.parser import HTMLParser
 except ImportError:
-    from HTMLParser import HTMLParser
+    from html.parser import HTMLParser
```
jan-cerny: The `try` section does the same thing as the `except` section now. That isn't correct.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -426,7 +426,7 @@ def _init_after_data_fetch(self, wait_for):
             self._ds_checklists = self._content_handler.get_data_streams_checklists()
             add_ds_ids = GtkActionList()
             add_ds_ids.add_action(self._ds_store.clear)
-            for dstream in self._ds_checklists.iterkeys():
+            for dstream in list(self._ds_checklists.keys()):
```
jan-cerny: Why is the `list()` function here? In python3, `.keys()` method returns a new view of the dictionary’s keys, which you can iterate over without any further conversions.

org_fedora_oscap/rule_handling.py
```diff
@@ -278,15 +278,15 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+        for part_rule in list(self._rules.values()):
             messages += part_rule.eval_rules(ksdata, storage, report_only)
 
         return messages
 
     def revert_changes(self, ksdata, storage):
         """:see: RuleHandler.revert_changes"""
 
-        for part_rule in self._rules.itervalues():
+        for part_rule in list(self._rules.values()):
```
jan-cerny: Same for `.values()`

org_fedora_oscap/gui/spokes/oscap.py
redhatrises: @jan-cerny this was more of a python 2 to python 3.6+ jump which I believe 3.6 uses/prefers `list(d.keys())` instead of just plain `d.keys()`. Correct me if I am wrong and misunderstanding something here.

org_fedora_oscap/rule_handling.py
redhatrises: See above.

org_fedora_oscap/content_handling.py
redhatrises: @jan-cerny fixed. Apparently wasn't paying attention.

org_fedora_oscap/gui/spokes/oscap.py
redhatrises: @jan-cerny I revisted this. What happened was that I had already fixed it correctly, then got distraced and fixed it again to be wrong. Fixed correctly now.

dahaic:
Jenkins test this please.

PR #48, opened 2017-09-26
When using content from SSG, use the DataStream by default

### LINKED ISSUE #41: Use SSG datastreams instead of SSG XCCDFs
When user customizes SSG profiles in SCAP Workbench, he customizes datastreams. When he saves tailoring file, he the file references the datastream. That's a complication when he will want to use it in OSCAP Anacoda Addon via kickstart, because OSCAP Anaconda Addon uses XCCDFs instead, so the  the original profile will not be found, and empty profile will be applied as a result.

yuumasato:
When loading content from SCAP Security Guide, we should use content from the DataStream, not from XCCDF file.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1465402
Also fixes #41

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -52,12 +52,12 @@ INSTALLATION_CONTENT_DIR = "/tmp/openscap_data/"
 TARGET_CONTENT_DIR = "/root/openscap_data/"
 
 SSG_DIR = "/usr/share/xml/scap/ssg/content/"
-SSG_XCCDF = "ssg-rhel7-xccdf.xml"
+SSG_SDS = "ssg-rhel7-ds.xml"
 if constants.shortProductName != 'anaconda':
     if constants.shortProductName == 'fedora':
-        SSG_XCCDF = "ssg-fedora-xccdf.xml"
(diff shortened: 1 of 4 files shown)


org_fedora_oscap/common.py
```diff
@@ -52,12 +52,12 @@
 TARGET_CONTENT_DIR = "/root/openscap_data/"
 
 SSG_DIR = "/usr/share/xml/scap/ssg/content/"
-SSG_XCCDF = "ssg-rhel7-xccdf.xml"
+SSG_SDS = "ssg-rhel7-ds.xml"
```
redhatrises: Please make this SSG_CONTENT

redhatrises:
@yuumasato Tests are still broken because of this change.

org_fedora_oscap/common.py
yuumasato: Oops.. will change

yuumasato:
Yeah, I'll fix the tests and try to add some more.

dahaic:
Thank you @yuumasato . I have just tried the installation with updated OAA, and this change successfully fixed the issue reported by BZ.

PR #49, opened 2017-10-24
RHBZ#1452667: Catch Anaconda Remediations parsing errors

jan-cerny:
The parser won't exit if it encounters something unexpected
in Ananconda remediations. Instead it will report a warning to log.

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -46,22 +46,36 @@ def parse_csv(option, opt_str, value, parser):
             parser.values.ensure_value(option.dest, []).append(item)
 
 
-PART_RULE_PARSER = optparse.OptionParser()
+class ModifiedOptionParserException(Exception):
+    """Exception to be raised by ModifiedOptionParser."""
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/rule_handling.py
```diff
@@ -170,10 +184,8 @@ def new_rule(self, rule):
         first_word = rule.split(None, 1)[0]
         try:
             actions[first_word](rule)
-        except KeyError:
```
yuumasato: We still need to catch `KeyError`. This catch is about not finding the rule in `actions` dictionary.

redhatrises: @jan-cerny per @yuumasato. Let's make sure to check for `KeyError`.

jan-cerny: thank you

jan-cerny: It will be here: https://github.com/OpenSCAP/oscap-anaconda-addon/pull/51

Issue #52, opened 2017-10-27
Method PasswdRules.eval_rules() ignores report_only parameter


jan-cerny:
In
https://github.com/OpenSCAP/oscap-anaconda-addon/blob/rhel7-branch/org_fedora_oscap/rule_handling.py#L444
`report_only` parameter is ignored.

It looks to me that it should change the values of variables only if `report_only` is `True`.

Could be related to https://bugzilla.redhat.com/show_bug.cgi?id=1374181

redhatrises:
Closing per PR #53

PR #53, opened 2017-11-07
Fix password rules cleaning

jan-cerny:
During manual installation when profile affecting password was selected,
selection of another profile did not clear the password requirements.
The method revert_changes didn't work because the saved old password
length was overwritten each time when eval_rules was invoked.
Saved old password length should be overwritten only if the eval_rules
method is called with report_only set to False.
Fixes rhbz #1374181 and issue #52.

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -472,6 +472,9 @@ class PasswdRules(RuleHandler):
             else:
                 ret = []
 
+        if report_only:
+            return ret
+
         # set the policy in any case (so that a weaker password is not entered)
         pw_policy = ksdata.anaconda.pwpolicy.get_policy("root")
         if pw_policy is None:


jan-cerny:
@pep8speaks Your issues are unrelated to my PR.

mpreisler:
> Also I would prefer to have ret changed to something like returned_msgs, to make the method easier to understand.

`ret` is a pretty standard name for something that will be returned as the result of the function. returned_msgs implied that something else returned these msgs. at least for me.

Plus I think this is out of scope of this change.

jan-cerny:
@dahaic I'll look into the possibility of unit tests.

Although I can see that ret is a weird name, it is very often used as a convention for a variable that contains return value of the function. I don't think it is necessary to change it.

dahaic:
@jan-cerny you know what - I think it's ok for now. Add tests asap ;)

PR #55, opened 2017-11-14
Add unittest for report_only parameter

jan-cerny:
This test should be part of PR #53, but that was already merged.


mpreisler:
ACK

PR #57, opened 2017-12-20
Improve test for password rule when report only is True

yuumasato:
#### Description:

- Added asserts for other relevant values which should not be changed when report
only is True.
- Check there is no warnings in `mesages` returned by `eval_rules`.
- It seemed strange to set `self.rule_data._passwd_rules._orig_minlen` and assert it right after. So I added mock for `ksdata.anaconda.pwpolicy.get_policy()`.


dahaic:
Looks good, thanks for the patch :)

PR #61, opened 2018-02-01
Define translation domain in OSCAPSpoke

yuumasato:
UI doesn't show translations if translation domain is not defined.

Below is screenshot of UI with translations.
screenshot from 2018-02-01 20-39-08

### DIFF
--- a/org_fedora_oscap/gui/spokes/oscap.glade
+++ b/org_fedora_oscap/gui/spokes/oscap.glade
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.18.3 -->
-<interface>
+<interface domain="oscap-anaconda-addon">
   <requires lib="gtk+" version="3.0"/>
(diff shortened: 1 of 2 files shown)


dahaic:
Good stuff @yuumasato - can you also make `SECURITY POLICY` string on the hub translated? Should change `N_` to `_` (plus removal of the `N_` definition):
https://github.com/OpenSCAP/oscap-anaconda-addon/blob/rhel7-branch/org_fedora_oscap/gui/spokes/oscap.py#L169

yuumasato:
@dahaic Unfortunately suggested change didn't work. `SECURITY POLICY` is still untranslated in hub.

dahaic:
Ok, we have discussed it a bit more with @yuumasato , and the problem is more complex. Anaconda itself needs to translate it, but probably does not use our domain. (Parts shipped in Anaconda itself have Hub titles provided together with context `GUI|Spoke`, adding it for us had no effect though). Needs deeper look into the code and how Anaconda integrates plugins. Possibly cannot be fixed on our side.

Thanks @yuumasato for your work on this!

PR #63, opened 2018-03-27
Add new kdump --disable rule handling

redhatrises:
- Adds new `kdump --disable` handling that can be added to anaconda scripts
  to be able to disable kdump from OSCAP content

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -721,33 +721,31 @@ class KdumpRules(RuleHandler):
 
         if self._kdump_enabled == None:
             return []
-
-        if self._kdump_enabled == False:
+        elif self._kdump_enabled == False:
             msg = _("Kdump will be disabled on startup")
-        if self._kdump_enabled == True:
+        elif self._kdump_enabled == True:
             msg = _("Kdump will be enabled on startup")
 
         messages.append(RuleMessage(self.__class__,
                                     common.MESSAGE_TYPE_INFO, msg))
 
-        try:
-            if not report_only:
+        if not report_only:
+            try:
                 ksdata.addons.com_redhat_kdump.enabled = self._kdump_enabled
-        except AttributeError:
-            log.warning("com_redhat_kdump is not installed. "
-                        "Skipping kdump configuration")
+            except AttributeError:
+                log.warning("com_redhat_kdump is not installed. "
+                            "Skipping kdump configuration")
 
         return messages
 
     def revert_changes(self, ksdata, storage):
         """:see: RuleHander.revert_changes"""
 
-        try:
-            if ksdata.addons.com_redhat_kdump.enabled == False or \
-               self._kdump_enabled == False:
+        if self._kdump_enabled == False:
+            try:
                 ksdata.addons.com_redhat_kdump.enabled = True
-        except AttributeError:
-            log.warning("com_redhat_kdump is not installed. "
-                        "Skipping reverting kdump configuration")
+            except AttributeError:
+                log.warning("com_redhat_kdump is not installed. "
+                            "Skipping reverting kdump configuration")
 
         self._kdump_enabled = None
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -729,3 +729,5 @@ class KdumpRules(RuleHandler):
 
         if not self._kdump_enabled:
             ksdata.addons.com_redhat_kdump.enabled = True
+
+        self._kdump_enabled = True
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -81,7 +81,7 @@ BOOTLOADER_RULE_PARSER.add_option("--passwd", dest="passwd", action="store_true"
 
 KDUMP_RULE_PARSER = ModifiedOptionParser()
 KDUMP_RULE_PARSER.add_option("--disable", action="store_false",
-                             dest="enabled", help="Disable kdump")
+                             dest="enabled", default=True)
 
 
 class RuleHandler(object):
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -80,8 +80,10 @@ BOOTLOADER_RULE_PARSER.add_option("--passwd", dest="passwd", action="store_true"
                                   default=False)
 
 KDUMP_RULE_PARSER = ModifiedOptionParser()
+KDUMP_RULE_PARSER.add_option("--enable", action="store_true",
+                             dest="kdenabled", default=None)
 KDUMP_RULE_PARSER.add_option("--disable", action="store_false",
-                             dest="enabled", default=True)
+                             dest="kdenabled", default=None)
 
 
 class RuleHandler(object):
@@ -249,8 +251,7 @@ class RuleData(RuleHandler):
         args = shlex.split(rule)
         (opts, args) = KDUMP_RULE_PARSER.parse_args(args)
 
-        if not opts.enabled:
-            self._kdump_rules.kdump_enabled()
+        self._kdump_rules.kdump_enabled(opts.kdenabled)
 
     @property
     def passwd_rules(self):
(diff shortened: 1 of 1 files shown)


redhatrises:
Sample service_kdump_disabled.anaconda output:

````$ cat shared/fixes/anaconda/service_kdump_disabled.anaconda
# platform = multi_platform_rhel

kdump --disable````

File: org_fedora_oscap/rule_handling.py [hunk 1]
```diff
@@ -668,3 +682,50 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+                log.warning("com_redhat_kdump is not installed. "
+                            "Skipping disabling kdump")
+
+        return messages
+
+    def revert_changes(self, ksdata, storage):
+        """:see: RuleHander.revert_changes"""
+
+        if not self._kdump_enabled:
+            ksdata.addons.com_redhat_kdump.enabled = True
```
mpreisler: this looks suspicious to me, shouldn't we rather save the previous state in eval_rules and the apply it here if it's not None?

redhatrises: @mpreisler I am pretty sure that we are just reverting everything that we did in eval_rules. I don't think that we are saving states from eval_rules at all. However, `self._kdump_enabled` should be reset back to `True` here.

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,71 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+
+    def eval_rules(self, ksdata, storage, report_only=False):
+        """:see: RuleHandler.eval_rules"""
+
+        messages = []
+
+        if self._kdump_enabled == None:
+            return []
+
+        if self._kdump_enabled == False:
```
mpreisler: elif plz

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,71 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+            msg = _("Kdump will be disabled on startup")
+        if self._kdump_enabled == True:
```
mpreisler: same

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,71 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+            msg = _("Kdump will be enabled on startup")
+
+        messages.append(RuleMessage(self.__class__,
+                                    common.MESSAGE_TYPE_INFO, msg))
+
+        try:
+            if not report_only:
+                ksdata.addons.com_redhat_kdump.enabled = self._kdump_enabled
```
mpreisler: shouldn't the try except be inside the condition?

mpreisler: you can also do if "com_redhat_kdump" in ksdata.addons

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,71 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+
+        try:
+            if ksdata.addons.com_redhat_kdump.enabled == False or \
+               self._kdump_enabled == False:
+                ksdata.addons.com_redhat_kdump.enabled = True
+        except AttributeError:
+            log.warning("com_redhat_kdump is not installed. "
+                        "Skipping reverting kdump configuration")
+
+        self._kdump_enabled = None
```
mpreisler: _kdump_enabled is basically a tribool, right? unknown, true or false. this looks like it will only revert kdump disable but never kdump enable

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,71 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+                        "Skipping kdump configuration")
+
+        return messages
+
+    def revert_changes(self, ksdata, storage):
+        """:see: RuleHander.revert_changes"""
+
+        try:
+            if ksdata.addons.com_redhat_kdump.enabled == False or \
+               self._kdump_enabled == False:
```
mpreisler: I do not understand this condition. why are you checking both?

redhatrises: forgot to remove that

org_fedora_oscap/rule_handling.py
redhatrises: @mpreisler kinda by design as kdump is enabled by default.

mpreisler: we shouldn't rely on that in the code logic IMO, it's pretty easy to do this in a fully generic way that will work even if Fedora or RHEL decide to disable kdump by default

redhatrises: @mpreisler recommendations?

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,78 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+    def __init__(self):
+        """Constructor setting the initial value of attributes."""
+
+        self._kdump_enabled = None
+        self._kdump_default_enabled = None
+
+    def kdump_enabled(self, kdenabled):
+        """Enable or Disable Kdump"""
+
+        if kdenabled == True:
```
mpreisler: how about: `if kdenabled is not None: self._kdump_enabled = kdenabled`?

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,78 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+    def __str__(self):
+        """Standard method useful for debugging and testing."""
+
+        ret = "kdump"
+
+        if self._kdump_enabled == True:
+            ret += " --enable"
+
+        if self._kdump_enabled == False:
+            ret += " --disable"
```
mpreisler: None is not handled here, not sure if it's okay or not

redhatrises: @mpreisler that is fine. We don't care about `None` here.

org_fedora_oscap/rule_handling.py
redhatrises: @mpreisler fixed

org_fedora_oscap/rule_handling.py
dahaic: Not sure why not to use
```
if self._kdump_enabled:
  ret += " --enable"
else:
  ret += " --disable"

Comment (redhatrises) [reply org_fedora_oscap/rule_handling.py]: @dahaic because `--disable` or `--enable` may not be selected at all. This logic is handled the same way for other classes as well in `rule_handling.py`

Comment (mpreisler) [reply org_fedora_oscap/rule_handling.py]: @dahaic it's essentially a tribool

File: org_fedora_oscap/rule_handling.py (hunk 5 above)
Comment (yuumasato) [reply org_fedora_oscap/rule_handling.py]: I think the logic can be simpler:
If `self._kdump_enabled is not None`, just assign `ksdata.addons.com_redhat_kdump.enabled` the value of `self._kdump_default_enabled`

Comment (redhatrises) [reply org_fedora_oscap/rule_handling.py]: Thanks @yuumasato! Good catch. Made the changes.

File: org_fedora_oscap/rule_handling.py [hunk 9]
```diff
@@ -668,3 +683,72 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+
+        return ret
+
+    def eval_rules(self, ksdata, storage, report_only=False):
+        """:see: RuleHandler.eval_rules"""
+
+        messages = []
+
+        if self._kdump_enabled == None:
```
matejak: Please either change 
* all the `==` in this if..elif..else block to `is`, or
* just check for `None` using `is`, and shorten the subsequent checks - you can safely assume that `self._kdump_enabled is False` is the same as `not self._kdump_enabled` and ditto for the converse case.

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,72 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+            msg = _("Kdump will be disabled on startup")
+        elif self._kdump_enabled == True:
+            msg = _("Kdump will be enabled on startup")
+
+        messages.append(RuleMessage(self.__class__,
+                                    common.MESSAGE_TYPE_INFO, msg))
+
+        if not report_only:
+            try:
+                if self._kdump_default_enabled == None:
```
matejak: Pls change `== None` to `is None`.

org_fedora_oscap/rule_handling.py
```diff
@@ -668,3 +683,72 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+class KdumpRules(RuleHandler):
+    """Simple class holding data from the rules affecting the kdump addon."""
+
+    def __init__(self):
+        """Constructor setting the initial value of attributes."""
+
+        self._kdump_enabled = None
+        self._kdump_default_enabled = None
+
+    def kdump_enabled(self, kdenabled):
```
matejak: This method doesn't seem to be doing anything. Or am I missing something?

matejak:
There are some formal issues with the code, but it is pretty close to merge as soon as they are resolved.

org_fedora_oscap/rule_handling.py
redhatrises: @matejak logic in the method was switched and incorrect. I corrected it now.

redhatrises:
Thanks for review @matejak! I fixed the issues.

matejak:
Nice one, thank you!

redhatrises:
Thanks again @matejak!

PR #65, opened 2018-04-20
Added PEP8 99-width config.

matejak:


mpreisler:
@openscap-ci add to whitelist

PR #68, opened 2018-06-12
Merge rhel7 branch into master.

matejak:
As tests are Python3-based, the following seems to make sense in this PR:

* merge the `rhel7` branch, then
* do the bare minimum to make it pass the tests.

It got somehow out of hand, as making OAA to work was very close to the mentioned bare minimum.


matejak:
Tests fail because of missing `pyanaconda` module.

mpreisler:
It seems to me that at least some of this should have landed in rhel7-branch. What is the reason for only shipping fixes in master? We will have to support the rhel7-branch for a long time.

create_update_image.sh
```diff
@@ -1,14 +1,187 @@
...
+	oscap-anaconda-addon
+"
+
+
+actions=(download_rpms install_rpms install_addon_from_repo create_image cleanup)
+
+
+download_rpms() {
+	mkdir -p "$rpmdir"
+	(cd "$rpmdir" && dnf download --arch x86_64,noarch $packages)
```
yuumasato: In the future the download may need to be improved to fetch packages from other releases.
I am using Fedora 27 so packages downloaded were fc27, but image installed was a Fedora 28.
It didn't matter and installation worked, I wonder if it can be a problem one day.

docs/manual/developer_guide.adoc
```diff
@@ -11,57 +11,92 @@ toc::[]
...
 
 **Required boot parameters:**
+
 ```
-inst.updates=http://192.168.122.1:8000/update.img
+inst.updates=http://gateway:8000/update.img
 ```
 
+Here, `gateway` is supposed to refer to your host machine.
+When setting the NAT network in the VM setup, `gateway` will be recognized correctly (at least on RHEL7).
```
yuumasato: I confirm `gateway` didn't work for Fedora 28.

docs/manual/developer_guide.adoc
```diff
@@ -11,57 +11,92 @@ toc::[]
...
+In other cases, input your IP address.
+
+Remember that you can also expose your custom-built SSG content to the insaller this way.
```
yuumasato: I suggest to say `provide` instead of `expose`.
And maybe explain quickly how to do it? Where to click on oscap-anaconda-addon.

matejak: I have rephrased this.

create_update_image.sh
matejak: I have added a releasever option, defaulting to 28. Good catch!

org_fedora_oscap/common.py
```diff
@@ -175,19 +175,21 @@ def _run_oscap_gen_fix(profile, fpath, template, ds_id="", xccdf_id="",
         raise OSCAPaddonError(msg)
 
     (stdout, stderr) = proc.communicate()
+    stdout_str = stdout.decode()
+    stderr_str = stderr.decode()
 
-    messages = re.findall(r'OpenSCAP Error:.*', stderr)
+    messages = re.findall(r'OpenSCAP Error:.*', stderr_str)
```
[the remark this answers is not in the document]
matejak: Great catch, thanks!

docs/manual/developer_guide.adoc
```diff
@@ -11,57 +11,92 @@ toc::[]
...
 
 - Manually
-- With PXE boot
+- With Network Install/PXE boot
 
 If you want to set it manually, you have to boot your machine into grub. Then you can change options (usually using "tab" key).
 
-If you use PXE boot you can pass requires parameters there. Advantage of this solution is that you will not need to change parameters during every boot.
+If you use Network Install/PXE boot you can pass requires parameters there - look for `kernel options`.
+Advantage of this solution is that you will not need to change parameters during every boot.
```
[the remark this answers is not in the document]
matejak: Let's leave it to another PR, as it affects both branches. I haven't used grub, so I can't tell.

PR #70, opened 2018-06-13
remove xccdf-path configuration option which can overwrite content-path

dardok:


redhatrises:
@openscap-ci add to whitelist

redhatrises:
@dardok xccdf-path is provided so that the content path can be overwritten.

dardok:
Oh, sorry - thought it was a typo, I accidentally used it with a -xccdf file *and* content-path set to a -ds and was confused as to why it ended up running the eval against the xccdf.  Rescinding...

redhatrises:
@dardok no worries :)

PR #69, opened 2018-06-13
fix use-after-close of zipfile and tfile

dardok:

PR #71, opened 2018-06-18
Zipfile handling fix backport

matejak:
Backport of #69


mpreisler:
I wish it was handled better with the `with` keyword and/or `try/finally` but that's not in the scope of this PR.

PR #74, opened 2018-07-11
Added pytest conversion of the content handling test.

matejak:
This PR adds a converted tests that were left out from #72.
The original file: https://github.com/OpenSCAP/oscap-anaconda-addon/blob/00726c1b18270c1c2da6b05f12101b2a7be7e0a8/tests/content_handling_test.py

For reference: [On pytest and exceptions](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions)

### DIFF
--- a/tests/test_content_handling.py
+++ b/tests/test_content_handling.py
@@ -1,0 +1,86 @@
+import os
+
+import pytest
+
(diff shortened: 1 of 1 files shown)


jan-cerny:
LGTM. Thanks.

PR #75, opened 2018-07-23
Merge branch 'rhel7-branch' into master.

matejak:
* Merged developer documentation.
* Merged conversion of tests to the `pytest` framework.


matejak:
@openscap-ci jenkins test this please.

jan-cerny:
Shouldn't we just merge the branches instead of creating a PR?

matejak:
@openscap-ci jenkins test this please.

matejak:
@openscap-ci jenkins test this please.

matejak:
@openscap-ci jenkins test this please.

matejak:
Waits for #76 to be resolved. Then, the PR needs to be updated with the appropriate .spec file.

jan-cerny:
@matejak #76 has been resolved :smiley_cat:

docs/manual/developer_guide.adoc
```diff
@@ -73,13 +73,14 @@ Advantage of this solution is that you will not need to change parameters during
...
 inst.updates=http://gateway:8000/update.img
-```
+----
 
 Here, `gateway` is supposed to refer to your host machine.
-When setting the NAT network in the VM setup, `gateway` will be recognized correctly (at least on RHEL7).
-In other cases, input your IP address.
+When setting the NAT network in the VM setup, `gateway` will be recognized correctly.
+This works if the `dnsmasq` is set up correctly, which is the case on RHEL systems, but not on Fedora.
+In other cases, input the IP address of the VM.
```
matusmarhefka: This is a little bit confusing. The only way I got it to work on Fedora was by selecting "network install" method and I used the address of my virtual bridge interface (`virbr0`) instead of `gateway` (as by default the python3 http.server binds to all interfaces). Then in "Network selection" part I selected the "Specify shared device name" and I put `virbr0` there. Maybe we can document this approach for Fedora use case.

docs/manual/developer_guide.adoc
```diff
@@ -103,16 +104,16 @@ There is also an official https://fedoraproject.org/wiki/How_to_debug_installati
...
 
-```
+----
 anaconda
 openscap-python
 python-cpio
 python-mock
-python-nose
-```
+python-pytest
```
matusmarhefka: IMO we should explicitly state `python3` for these packages. There is also `openscap-python3`.

matusmarhefka:
@matejak I have an issue in GUI installation, after fetching `ssg-fedora-ds.xml`, selecting a security policy and starting the installation process there is an error stating that `openscap` and `openscap-scanner` packages should be installed and that they do not exist. This happens for Fedora 28 installation with `update.img` created using `./create_update_image.sh download_rpms` (I checked the image using `lsinitrd update.img` and it contains openscap library and scanner and the `rpm` folder contains the correct rpms for fc28). During the installation I can see that `update.img` was downloaded from the python http server and I can also see it in Anaconda logs. Am I doing something wrong?

docs/manual/developer_guide.adoc
matejak: Yes, I have rephrased this.

oscap-anaconda-addon.spec
```diff
@@ -1,45 +1,31 @@
...
-URL:            https://github.com/OpenSCAP/oscap-anaconda-addon.git
+URL:            https://github.com/OpenSCAP/oscap-anaconda-addon
 
 # This is a Red Hat maintained package which is specific to
 # our distribution.
 #
 # The source is thus available only from within this SRPM
 # or via direct git checkout:
-# git clone https://github.com/OpenSCAP/oscap-anaconda-addon.git
+# git clone git://git.fedorahosted.org/oscap-anaconda-addon.git
```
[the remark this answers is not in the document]
matejak: Fixed, this has been caused by using whole another `.spec` file.

docs/manual/developer_guide.adoc
matejak: I have explicitly specified `python3-`, but `pycurl` is missing for both branches, so I consider it out of scope of this PR. See #78.

docs/manual/developer_guide.adoc
```diff
@@ -35,25 +35,25 @@ If some time has passed since you have created the image, you typically want to
...
+----
 ./create_update_image.sh download_rpms
-```
+----
 
 If you want to see what was packed, you can extract the image.
 
-```
+----
 gunzip -c update.img | cpio -id
```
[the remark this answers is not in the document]
matejak: This is missing for both branches, so I consider it out of scope of this PR. See #78.

matejak:
@matusmarhefka Concerning your run-time error, you have not done anything wrong besides going for wrong compose - e.g. [Everything](https://dl.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os/) works, but [Workstation](http://mirror.karneval.cz/pub/linux/fedora/linux/development/28/Workstation/x86_64/os/) doesn't.

matusmarhefka:
ACK.

PR #76, opened 2018-07-24
Replaced the spec file with spec file closer to the RHEL7 spec.

matejak:
There feeling in our team is that it is beneficial to have the upstream `.spec` files more similar to what we actually use downstream.
@redhatrises AFAIK, you were the person behind having `.spec` files to take part in PR gating, do you still like that idea? The .spec file in master and rhel7 are going to diverge.

### DIFF
--- a/oscap-anaconda-addon.spec
+++ b/oscap-anaconda-addon.spec
@@ -1,10 +1,4 @@
-%if 0%{?fedora} >= 25
-%define _py python3
-%endif
-
-%if 0%{?rhel} || 0%{?fedora} < 25
-%define _py python
-%endif
+%global _default_patch_flags --no-backup-if-mismatch
 
 Name:           oscap-anaconda-addon
 Version:        0.9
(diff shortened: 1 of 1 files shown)


redhatrises:
@matejak with the changes can this still be built and run on the latest version of Fedora and Anaconda? If not, I say that we should NOT make the spec files the same. In fact, I was in an internal meeting yesterday asking why OAA is not the latest in Fedora and when will it be updated to the latest in Fedora (which I have asked multiple times myself).

Some of these changes remove the handling for py2 and py3 (which may not be necessary anymore) as well as comment out `%check` which I am not sure that that is a good idea either.

matejak:
@redhatrises This is the RHEL7 branch - i.e. the OAA 0.8, perhaps 0.9 versions. The OAA which works with modern Anaconda in Fedora 28 and uses Python3 is OAA 1.0. OAA 1.0 is available in Fedora 28, but not in F27 as it wouldn't work in Fedora<28 due to Anaconda changes.
My view is that we either have corresponding spec files in each branch (i.e. RHEL7 and Fedora 28 without patches), or we get rid of those spec files in upstream completely. Changes to those spec files that you have noticed are caused by the effort to have as similar things upstream and downstream as possible.
Commenting tests out was a glitch.

redhatrises:
> @redhatrises This is the RHEL7 branch - i.e. the OAA 0.8, perhaps 0.9 versions. 

🤦‍♂️ Thanks @matejak. Looks good.

cipherboy:
This looks sane. Since @redhatrises approved of including this in the RHEL7 branch, I'll merge this. :)

cipherboy:
Thanks @matejak !

PR #81, opened 2018-10-09
Updated the code to work with current devel Anaconda.

matejak:
The Anaconda API between the `F28` and current `master` branch differs, so the OAA code also needs to be updated.

What has been done:
* The `ksdata.firewall` interface has been updated to use the `firewall_proxy` interface. There is a slight nuisance here that the proxy interface supports get/set methods, but not set operations s.a. add, difference, etc., so the code doesn't do a 100% right thing. I have decided to leave it like it is for the time being, as we are under time pressure, and the firewall settings are not used in our security content.
* The `ksdata.rootpw` has been replaced by `users_proxy` interface.
* The `storage.bootloader` has been replaced by the `bootloader_proxy` interface. Note: The interface seems to have expanded, so we can aim to set the bootloader password in the future.

What has not been done:
* The `storage.mountpoints` interface conversion (As confirmed by Anaconda people, this is OK, as `storage` in this context points to `blivet`, which has the same API as we are used to).
* The `ksdata.packages` interface conversion (As confirmed by Anaconda people, this is OK).
* The `ksdata.addons.com_redhat_kdump` interface that queries settings of the Kdump addon (As confirmed by Anaconda people, this is OK, we may have to check with Kdump people concerning their API, but it is extremely unlikely that they change it).

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -26,7 +26,13 @@ Module with various classes for handling pre-installation rules.
 import optparse
 import shlex
 import logging
+
 from pyanaconda.pwpolicy import F22_PwPolicyData
+from pyanaconda.core.constants import (
+    FIREWALL_ENABLED, FIREWALL_DISABLED, FIREWALL_USE_SYSTEM_DEFAULTS)
+from pyanaconda.modules.common.constants.objects import FIREWALL, BOOTLOADER
+from pyanaconda.modules.common.constants.services import NETWORK, STORAGE, USERS
+
 from org_fedora_oscap import common
 from org_fedora_oscap.common import OSCAPaddonError, RuleMessage
 
@@ -496,7 +502,10 @@ class PasswdRules(RuleHandler):
             return []
 
         ret = []
-        if not ksdata.rootpw.password:
+
+        users_proxy = USERS.get_proxy()
+
+        if not users_proxy.IsRootPasswordSet:
             # root password was not set
 
             msg = _("make sure to create password with minimal length of %d "
@@ -505,12 +514,12 @@ class PasswdRules(RuleHandler):
                                common.MESSAGE_TYPE_WARNING, msg)]
         else:
             # root password set
-            if ksdata.rootpw.isCrypted:
+            if users_proxy.IsRootPasswordCrypted:
                 msg = _("cannot check root password length (password is crypted)")
                 log.warning("cannot check root password length (password is crypted)")
                 return [RuleMessage(self.__class__,
                                     common.MESSAGE_TYPE_WARNING, msg)]
-            elif len(ksdata.rootpw.password) < self._minlen:
+            elif len(users_proxy.RootPassword) < self._minlen:
                 # too short
                 msg = _("root password is too short, a longer one with at "
                         "least %d characters is required") % self._minlen
@@ -705,10 +714,13 @@ class BootloaderRules(RuleHandler):
     def eval_rules(self, ksdata, storage, report_only=False):
         """:see: RuleHandler.eval_rules"""
 
-        if self._require_password and not storage.bootloader.password:
-            # Anaconda doesn't provide a way to set bootloader password, so
-            # users cannot do much about that --> we shouldn't stop the
-            # installation, should we?
+        bootloader_proxy = STORAGE.get_proxy(BOOTLOADER)
+
+        if self._require_password and not bootloader_proxy.password_is_set:
+            # TODO: Anaconda provides a way to set bootloader password:
+            # bootloader_proxy.set_password(...)
+            # We don't support setting the bootloader password yet,
+            # but we shouldn't stop the installation, just because of that.
             return [RuleMessage(self.__class__, common.MESSAGE_TYPE_WARNING,
                                 "boot loader password not set up")]
         else:
@@ -802,8 +814,13 @@ class FirewallRules(RuleHandler):
         self._added_trusts = set()
         self._removed_svcs = set()
 
+        self._new_services_to_add = set()
+        self._new_ports_to_add = set()
+        self._new_trusts_to_add = set()
+        self._new_services_to_remove = set()
+
         self._firewall_enabled = None
-        self._firewall_default_enabled = None
+        self._firewall_default_state = None
 
     def add_services(self, services):
         """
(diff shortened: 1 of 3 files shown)


org_fedora_oscap/rule_handling.py
```diff
@@ -895,25 +907,26 @@ def __str__(self):
     def eval_rules(self, ksdata, storage, report_only=False):
         """:see: RuleHandler.eval_rules"""
 
+        firewall_proxy = NETWORK.get_proxy(FIREWALL)
         messages = []
 
         if self._firewall_default_enabled is None:
             # firewall default startup setting
-            self._firewall_default_enabled = ksdata.firewall.enabled
+            self._firewall_default_enabled = firewall_proxy.FirewallMode
```
yuumasato: I would suggest to call this variable `_firewall_detault_state`, the current name is confusing.

org_fedora_oscap/rule_handling.py
```diff
@@ -938,48 +951,51 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
             if not report_only:
                 self._added_svcs.add(svc)
-                ksdata.firewall.services.append(svc)
 
             msg = _("service '%s' has been added to the list of services to be "
                     "added to the firewall" % svc)
             messages.append(RuleMessage(self.__class__,
                                         common.MESSAGE_TYPE_INFO, msg))
+        if not report_only:
+            firewall_proxy.SetEnabledServices(self._add_svcs)
```
yuumasato: I couldn't exactly find in documentation, but from [source code](https://github.com/rhinstaller/anaconda/tree/master/pyanaconda/modules/network/firewall), it looks like the new interface doesn't append, it just sets the new list of services.

So this should probably contain both the newly enabled and already enabled services.

org_fedora_oscap/rule_handling.py
```diff
@@ -938,48 +951,51 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
             if not report_only:
                 self._added_ports.add(port)
-                ksdata.firewall.ports.append(port)
 
             msg = _("port '%s' has been added to the list of ports to be "
                     "added to the firewall" % port)
             messages.append(RuleMessage(self.__class__,
                                         common.MESSAGE_TYPE_INFO, msg))
+        if not report_only:
+            firewall_proxy.SetEnabledPorts(self._add_ports)
```
yuumasato: The same comment as in `SetEnabledServices`.

org_fedora_oscap/rule_handling.py
```diff
@@ -938,48 +951,51 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
             if not report_only:
                 self._added_trusts.add(trust)
-                ksdata.firewall.trusts.append(trust)
 
             msg = _("trust '%s' has been added to the list of trusts to be "
                     "added to the firewall" % trust)
             messages.append(RuleMessage(self.__class__,
                                         common.MESSAGE_TYPE_INFO, msg))
+        if not report_only:
+            firewall_proxy.SetTrusts(self._add_trusts)
```
yuumasato: The same comment as in `SetEnabledServices`.

org_fedora_oscap/rule_handling.py
```diff
@@ -990,28 +1006,30 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
             if not report_only:
                 self._removed_svcs.add(svc)
-                ksdata.firewall.remove_services.append(svc)
 
             msg = _("service '%s' has been added to the list of services to be "
                     "removed from the firewall" % svc)
             messages.append(RuleMessage(self.__class__,
                                         common.MESSAGE_TYPE_INFO, msg))
+        if not report_only:
+            firewall_proxy.SetDisabledServices(self._removed_svcs)
```
yuumasato: The same comment as in `SetEnabledServices`.

org_fedora_oscap/rule_handling.py
jkonecny12: Yes the documentation is outdated for the current code, sorry for that. And again yes, as you said this should set a list not append.

org_fedora_oscap/rule_handling.py
yuumasato: Or `_firewall_default_mode`.

matejak:
I have rebased, and incorporated fixes for flaws pointed by @yuumasato . The code needs a more serious refactoring, but that is out of scope for this PR.

org_fedora_oscap/rule_handling.py
```diff
@@ -937,49 +950,55 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
             if not report_only:
                 self._added_svcs.add(svc)
-                ksdata.firewall.services.append(svc)
 
             msg = _("service '%s' has been added to the list of services to be "
                     "added to the firewall" % svc)
             messages.append(RuleMessage(self.__class__,
                                         common.MESSAGE_TYPE_INFO, msg))
+        if not report_only:
+            all_services = list(self._add_svcs.union(set(firewall_proxy.EnabledServices)))
```
yuumasato: Would it be better to put together add `self._added_svcs` to `firewall_proxy.EnabledServices`? We know they will be disjoint sets.

matejak: I would say that it's the same.

matejak:
@yuumasato I have incorporated changes to the revert functionality.

org_fedora_oscap/rule_handling.py
```diff
@@ -990,52 +1017,56 @@ def eval_rules(self, ksdata, storage, report_only=False):
...
+        all_ports = firewall_proxy.EnabledPorts
+        orig_ports = set(all_ports).difference(self._new_ports_to_add)
+        firewall_proxy.SetEnabledPorts(list(orig_ports))
 
         # remove all trusts this handler added
-        for trust in self._added_trusts:
-            if trust in ksdata.firewall.trusts:
-                ksdata.firewall.trusts.remove(trust)
+        all_trusts = firewall_proxy.Trusts
+        orig_trusts = set(all_trusts).difference(self._new_ports_to_add)
```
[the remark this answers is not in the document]
matejak: good catch! Fixed.

yuumasato:
@matejak Thanks for these updates!

PR #88, opened 2019-01-03
Fix typos in developer guide documentation.

ggbecker:


ggbecker:
@openscap-ci add to whitelist

jan-cerny:
The CI fail is not related to your changes. The patch looks great, but the PR needs to be opened against `rhel7-branch` branch because these typos are also present there. That way we will eventually fix all the branches. Could you switch the target branch of the PR to  `rhel7-branch`? Thanks.

ggbecker:
> The CI fail is not related to your changes. The patch looks great, but the PR needs to be opened against `rhel7-branch` branch because these typos are also present there. That way we will eventually fix all the branches. Could you switch the target branch of the PR to `rhel7-branch`? Thanks.

I think simply changing the branch isn't the best way to do it because of different commits between branches. I'll create another one based on rhel7-branch.

PR #89, opened 2019-01-04
Fix typos in developer guide documentation.

ggbecker:


ggbecker:
@openscap-ci add to whitelist

PR #90, opened 2019-01-04
Remove outdated Makefile help usage and update development guide.

ggbecker:
Update development guide with available Makefile commands.

Add a warning when intltool is not installed in the system similarly to the one for when Zanata is not installed either.


ggbecker:
@openscap-ci add to whitelist

jan-cerny:
@openscap-ci test this please

PR #92, opened 2019-01-07
Remove translations from upstream CI 

jan-cerny:
Pulling from Zanata every time there is a PR update is useless waste,
so tests for the upstream CI should be modified.

The run without pulling from Zanata takes 4 seconds, with pulling from Zanata it takes ~ 2 minutes and it creates traffic.

The rpmbuild is a part of CI, therefore we need to modify the spec file that is used.

### DIFF
--- a/oscap-anaconda-addon.spec
+++ b/oscap-anaconda-addon.spec
@@ -44,9 +44,8 @@ make unittest
 
 %install
 make install DESTDIR=%{buildroot}
-%find_lang %{name}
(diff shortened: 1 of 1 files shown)


yuumasato:
LGTM.

jan-cerny:
We will probably need to revert this and handle the problem in Makefiles but not int spec file, because in fact the Makefile contains commands to download data from Zanata.

PR #93, opened 2019-01-10
Update unit tests after upgrading code to latest Anaconda code

jan-cerny:
In some of the `rule_handling` methods that handle kickstart rules we recently started to use new Anaconda API that uses DBUS proxy. See https://github.com/OpenSCAP/oscap-anaconda-addon/pull/81

However, after this change our upstream tests started to fail.  The problem with
upstream tests was that they don't replace the DBus proxy by a mock object and also that they mocked the old classes that aren't used in the SUT.

This PR updates our unit test by mocking the DBus proxy. We keep the mock object that is returned by `DBus.get_proxy`, modify its values and then run the code that uses these values. The test update is inspired by:
https://github.com/rhinstaller/anaconda/blob/master/tests/nosetests/pyanaconda_tests/partitioning_test.py#L48

### DIFF
--- a/tests/test_rule_handling.py
+++ b/tests/test_rule_handling.py
@@ -412,6 +412,7 @@ def test_evaluation_passwd_minlen_short_passwd(
     # doing changes --> password should not be cleared
     assert password_proxy_mock.RootPassword == "aaaa"
 
+
 @mock.patch("pyanaconda.dbus.DBus.get_proxy")
 def test_evaluation_passwd_minlen_short_passwd_report_only(
         proxy_getter, rule_data, ksdata_mock, storage_mock):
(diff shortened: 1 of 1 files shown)


jan-cerny:
@openscap-ci test this please

jan-cerny:
The PRs which are opened against rhel8-branch shouldn't run CI on RHEL 7 machines. We need to update the Jenkins configuration accordingly.

jan-cerny:
@openscap-ci test this please

matejak:
Good job, let's hope that the Anaconda project will provide some usable prepared mocks in the future, so our tests will be more elegant and powerful at the same time.

PR #94, opened 2019-01-14
Revert "Remove translations from upstream CI "

jan-cerny:
Reverts OpenSCAP/oscap-anaconda-addon#92


jan-cerny:
We need to revert #92 and handle the problem in Makefiles but not in the spec file, because in fact the Makefile contains commands to download data from Zanata.

PR #95, opened 2019-01-14
Remove spec file from upstream repository

jan-cerny:
Consumers of the spec file should refer to Fedora packaging repository
where the packaging information can be found:
https://src.fedoraproject.org/rpms/oscap-anaconda-addon/blob/master/f/oscap-anaconda-addon.spec
You can also refer to the specfiles shipped in Red Hat Enterprise Linux.
We learned through history that maintaining the same spec file in
multiple repositories was a tedious job. There were multiple reasons why
the specfiles diverged from each other. We already removed upstream spec
files from OpenSCAP GitHub repository in past.


yuumasato:
LGTM.
But one information which is in the specfile which is not anywhere is the required anaconda version, I believe we should keep track of it somewhere.

yuumasato:
Ack, I'll add the info to README file, there is no need to hold this PR.

PR #96, opened 2019-01-14
Add compatible version of anaconda to rhel7 README

yuumasato:
Add Anaconda version to which this add-on is compatible.


yuumasato:
After this one is merged I will update other branches as well.

PR #103, opened 2019-01-17
Updated the code to use the up-to-date Anaconda API.

matejak:
Tested on a VM and it worked.

### DIFF
--- a/org_fedora_oscap/gui/spokes/oscap.py
+++ b/org_fedora_oscap/gui/spokes/oscap.py
@@ -38,6 +38,8 @@ from pyanaconda.ui.gui.utils import set_treeview_selection, fire_gtk_action
 from pyanaconda.ui.categories.system import SystemCategory
 from pykickstart.errors import KickstartValueError
 
+from pyanaconda.modules.common.constants.services import USERS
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -650,26 +652,35 @@ def _update_message_store(self, report_only=False):
...
+
             if not report_only:
-                self.__old_root_pw = self.data.rootpw.password
+                users_proxy = USERS.get_proxy()
+
+                self.__old_root_pw = users_proxy.RootPassword
                 self.data.rootpw.password = None
-                self.__old_root_pw_seen = self.data.rootpw.seen
+                self.__old_root_pw_seen = users_proxy.IsRootpwKickstarted
                 self.data.rootpw.seen = False
```
poncovka, line 674: I don't think this fix is working correctly. You should call `users_proxy.SetRootPassword` and `users_proxy.SetRootpwKickstarted` as you do bellow. The `rootpw` is attribute is available, but useless in RHEL 8. It is no longer available in RHEL 9.

PR #104, opened 2019-02-07
Hack hub title to show translated.

dahaic:
This is workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1673071

Currently, Anaconda tries to translate the plugin title, but being in different translation domain prevents it to succeed. This way, we provide already translated string.

Thanks @poncovka!


yuumasato:
Great, the patch works for both RHEL7 and RHEL8.

PR #107, opened 2019-08-21
Do not use capitals for spoke title.

rvykydal:
Resolves: rhbz#1744185

To be consistent with other spokes.


jan-cerny:
@openscap-ci ok to test

matejak:
Tests failures are expected, so we can merge this change.

PR #119, opened 2020-06-17
Fixed handling of languages + oscap title update.

matejak:
This PR combines two things:

- Allow generation of update image with or without local translation (.mo) files. This involves following changes:
  - in the Makefile, add `RPM_BUILD_ROOT=$(DESTDIR)` to install language files to the correct location.
  - in the Makefile, add a `DEFAULT_INSTALL_OF_PO_FILES` variable that is a toggle install/don't install language files. Install them by default.
  - update the `create_update_image.sh` script, so it accepts the `-l` option to include local translations.
  - regenerate that script using newer Argbash.

- Change the spoke title to not use capital letters, and update the pot file accordingly.
- Revert a change in a string, where change from URL to URI causes mess in translations.

For some reason, the spoke title translation doesn't work any more as it did for 8.2. This changes the spoke title that also serves as an input for translations. As those translations are not updated, if you would like to see the spoke translated, you have to perform following steps:
1. Choose a language, and open its `po/lang.po` file.
2. Search for `SECURITY POLICY` and change occurrences to `Security Policy`. There should be two such occurrences.
3. Rebuild the `.mo` file by invoking `msgfmt po/lang.po -o po/lang.mo`
4. Rebuild the image with the `-l` option.
5. Choose that language during the installation.


po/oscap-anaconda-addon.pot
```diff
@@ -274,11 +274,6 @@ msgstr ""
 msgid "_Use SCAP Security Guide"
 msgstr ""
 
-#: tmp/oscap.glade.h:12
-msgid ""
-"No content found. Please enter data stream content or archive URI below:"
```
ggbecker, line 279: it took me a few minutes to understand why this is being removed, but now I get it. :)

Before it was `URL` than changed to `URI` than changed back to `URL`

matejak: A more elaborate explanation is part of the commit message.

ggbecker: I saw that, but only after going through some `blaming`.

PR #120, opened 2020-06-17
Fix the mock function for DBus.get_proxy

poncovka:
The function should be able to accept other arguments.


matejak:
Just checking, is this inded intended for the "next Fedora" branch?

poncovka:
@matejak, yes, this change is required for the current rawhide. I don't expect that it will be required in RHEL 8.

matejak:
Thanks, LGTM, therefore merging.

PR #125, opened 2020-08-27
Update the language tooling from Zanata to Weblate for RHEL8

matejak:
Code is heavily based on contents of
https://github.com/rhinstaller/anaconda/blob/master/Makefile.am

Changes:

- Weblate is Github-based, so git and file operations are used instead of a CLI client.
- There is no push-pot, but update-pot, which leaves the latest step (git push) up to you.

Analogous PRs will have to be ported to other branches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,14 @@ EXCLUDES = \
 	*~ \
 	*.pyc
 
-ZANATA_PULL_ARGS = --transdir po/
-ZANATA_PUSH_ARGS = --srcdir po/ --push-type source --force
+L10N_REPO_RELATIVE_PATH ?= OpenSCAP/oscap-anaconda-addon-l10n.git
+L10N_REPOSITORY ?= https://github.com/$(L10N_REPO_RELATIVE_PATH)
+L10N_REPOSITORY_RW ?= git@github.com:$(L10N_REPO_RELATIVE_PATH)
+# Branch used in anaconda-l10n repository.
+# This should be master all the time, unless you are testing translation PRs.
+GIT_L10N_BRANCH ?= master
+# The base branch, used to pair code with translations
+OAA_PARENT_BRANCH ?= rhel8-branch
 
 all:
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -73,13 +79,47 @@ dist:
...
+		popd ; \
+		echo "Pot file is up to date" ; \
+		rm -rf $$TEMP_DIR ; \
+	else \
+		git add ./$(POTFILE_BASENAME) && \
+		git commit -m "Update $(POTFILE_BASENAME)" && \
+		popd && \
+		echo "Pot file updated for the localization repository $(L10N_REPOSITORY)" && \
+		echo "Please confirm changes (git diff HEAD~1) and push:" && \
+		echo "$$TEMP_DIR" ; \
```
jan-cerny, line 121: Do I understand it well that the user should cd to the TEMP_DIR and run git diff, git add, git commit and git push there?

Makefile
```diff
@@ -73,13 +79,47 @@ dist:
...
+# - copy pot file to this repository
+# - check if pot file is changed (ignore the POT-Creation-Date otherwise it's always changed)
+# - if not changed:
+#   - remove cloned repository
+# - if changed:
+#   - add pot file
+#   - commit pot file
+#   - tell user to verify this file and push to the remote from the temp dir
+POTFILE_BASENAME = oscap-anaconda-addon.pot
+update-pot:
```
jan-cerny, line 104: It seems to require `intltool-extract`. Can we document this somewhere?

```
[jcerny@thinkpad oscap-anaconda-addon{pr/125}]$ make update-pot
make -C po potfile
make[1]: Entering directory '/home/jcerny/work/git/oscap-anaconda-addon/po'
rm -rf tmp/
You may not have the intltool-extract installed, don't be surprised if the operation fails.
for f in ../org_fedora_oscap/gui/spokes/oscap.glade; do \
        intltool-extract --type=gettext/glade -l $f ;\
done
/bin/sh: line 1: intltool-extract: command not found
make[1]: *** [Makefile:45: glade-po] Error 127
make[1]: Leaving directory '/home/jcerny/work/git/oscap-anaconda-addon/po'
make: *** [Makefile:105: update-pot] Error 2
[jcerny@thinkpad oscap-anaconda-addon{pr/125}]$ 
```

jan-cerny, line 104: When am I supposed to run `make update-pot`? Is it run during the release?

jan-cerny (commented):
The code looks fine. I have some comments. I'm sorry if the answers are obvious, but I'm not much familiar with this process.

Makefile
matejak: That's the interesting part. Ideally, that should happen after every PR merge, but it was never done that way even before. So it if fine when we do it upon release.

Makefile
matejak: Try to do it yourself. From what I see in the code, the temp dir contains a repository that got the file added and committed, so the only thing left to do is to push.

PR #129, opened 2020-11-05
Replace "policy" with "profile"

redhatrises:
- OAA applies security profiles and not security policy. Security policy is more than what OAA supports, and the policy specifies what profile to use (e.g. stig, cis, etc.) and customization to make to those profiles. This fixes that misnomer.


matejak:
Have you considered proposing that change against `master`? A change of terminology is not really a typo, and although I agree that it is a change in right direction, I would keep the current name for RHEL8.

redhatrises:
> Have you considered proposing that change against `master`? A change of terminology is not really a typo, and although I agree that it is a change in right direction, I would keep the current name for RHEL8.

If you would prefer it under master, then sure.

matejak:
Yes, please, that would be great.

redhatrises:
See #130

PR #130, opened 2020-11-06
Replace "policy" with "profile"

redhatrises:
- OAA applies security profiles and not security policy. This fixes that misnomer.


matejak:
Wow, that was fast. LGTM.

PR #131, opened 2020-12-03
Remove the justUpdate argument

poncovka:
Call `hubQ.send_ready` without the `justUpdate` argument.
The support for this argument was removed.

**Depends on:** https://github.com/rhinstaller/anaconda/pull/3028


jan-cerny:
@poncovka thank you very much

PR #134, opened 2020-12-09
Fix some of the storage-related tests

poncovka:
There are some tests that use a mock of the local storage object, but this
object is no longer available and it is always None. Test a mock a DBus
proxy instead.

### DIFF
--- a/tests/test_rule_handling.py
+++ b/tests/test_rule_handling.py
@@ -121,7 +121,7 @@ def ksdata_mock():
 
 @pytest.fixture()
 def storage_mock():
-    return mock.Mock()
+    return None
 
 
 # monkeypatch is a predefined fixture that is used to perform per-test case changes in the test env.
@@ -682,7 +682,8 @@ def test_evaluation_various_rules(proxy_getter, rule_data, ksdata_mock, storage_
                  "package --add=firewalld", ]:
         rule_data.new_rule(rule)
 
-    storage_mock.mountpoints = dict()
+    ksdata_mock.packages.packageList = []
+    ksdata_mock.packages.excludedList = []
 
     messages = rule_data.eval_rules(ksdata_mock, storage_mock)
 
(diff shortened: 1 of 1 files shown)


tests/test_rule_handling.py
```diff
@@ -657,37 +656,52 @@ def test_evaluation_various_rules(proxy_getter, rule_data, ksdata_mock, storage_
...
-    storage_mock.mountpoints = dict()
 
     messages = rule_data.eval_rules(ksdata_mock, storage_mock)
 
     # mount point doesn't exist -> one message, nothing done
     assert len(messages) == 1
-    assert storage_mock.mountpoints == dict()
+
+    device_tree_mock = STORAGE.get_proxy(DEVICE_TREE)
+    device_tree_mock.SetDeviceMountOptions.assert_not_called()
```
matejak, line 666: Does this assert tell us anything? The device tree has just been created, so testing that a method of itself was not called doesn't look like something to be tested in the addon code.

matejak (changes requested):
Thanks for updating our tests. I suggest to make `/dev/sda1` a constant (what about `device_for_tmp`), so it is clear what is its role.

Out of scope of this PR, but could you please tell whether it makes sense:
It is interesting that `STORAGE.get_proxy(DEVICE_TREE)` returns a mock, would it be possible to return something more intelligent, so one could write `device_tree_mock.mount_options["/dev/sda1"] == "defaults"` or something generally more expressive?

tests/test_rule_handling.py
poncovka: The `proxy_getter` fixture makes sure that `STORAGE.get_proxy(DEVICE_TREE)` always returns the same mock object. Therefore, this mock object might be already used in the `eval_rules` method. I can move the definition of the `device_tree_mock` variable at the beginning of the test if it is too confusing.

poncovka:
> It is interesting that `STORAGE.get_proxy(DEVICE_TREE)` returns a mock, would it be possible to return something more intelligent, so one could write `device_tree_mock.mount_options["/dev/sda1"] == "defaults"` or something generally more expressive?

The mocks are initialized in `set_dbus_defaults`, where you can change the defaults. You are able to define the `mount_options` dictionary and change `GetDeviceMountOptions` and `SetDeviceMountOptions` to work with this directory. I do something similar in `test_revert_mount_options`. Sometimes it is necessary to implement the side effects, but I don't think that it is necessary for all tests. The important part is the DBus call.

matejak (approved):

matejak:
Thanks for the update, I have no objections against those changes, and nitpicks can be dealt with later.

PR #135, opened 2020-12-09
Use the DBus API of the Kdump add-on

poncovka:
Don't use kickstart data to work with the Kdump add-on. The data are not
available, because the add-on has been migrated on DBus. Use the DBus API
instead.

**Related to:** https://github.com/daveyoung/kdump-anaconda-addon/pull/29

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -37,7 +37,10 @@ import logging
 from collections import namedtuple
 import gettext
 from functools import wraps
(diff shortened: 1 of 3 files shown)


poncovka:
Again, I am not sure how to test it on Rawhide.

matejak:
Those PRs can't be properly tested due to the state of rawhide even without the addon.

matejak:
LGTM

PR #139, opened 2021-03-02
Introduce packit/source-git support

matejak:


TomasTomecek:
we don't invoke actions in shell by default so you need to wrap commands with `bash -c "..."` if you need it

PR #140, opened 2021-03-11
Merge the cpio module into the addon

matejak:
The Python cpio module was not properly maintained upstream.
The LGPL license allows to relicence the derivative work under GPL - see LGPL section 3 or http://www.gnu.org/licenses/gpl-faq.html#AllCompatibility


yuumasato (approved):
Thank you @matejak,  LGTM.

Approving but not merging for now as there are a few changes in master pending a walk through the test valley.

PR #141, opened 2021-04-23
Fix the rawhide package handling

matejak:
The "fedora-latest" Anaconda diverged from the "rhel-latest", so the master branch got a bump in the major version number, and the first difference between this and the previous major versions has landed.


poncovka (approved):
Looks good to me.

PR #143, opened 2021-05-21
Add packages

jan-cerny:
Starting from OpenSCAP 1.3.5, OpenSCAP depends on xmlsec1 library.
Therefore, OAA needs to have these packages present in the update
img.
.#


matejak:
Correct, thanks for the PR!

PR #144, opened 2021-05-24
Document reanaconda

jan-cerny:
It's an easier way of running a test VM for development and testing of OSCAP Anaconda Addon.

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -15,8 +15,7 @@ What do you need to test your changes?
 
 1. Clone addon repository & use required branch & change code.
 2. Build addon update image.
-3. Serve image using local HTTP server.
-4. Load system with update image.
+3. Run a VM using reanaconda.
 
 
 === Clone repository & use required branch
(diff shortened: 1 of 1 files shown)


matejak:
This is perfect, thank you very much! However, could you please rebase this and change the target to rhel8-branch? The RHEL7 one would be applicable too, but forward-porting this would then be quite a pain.

jan-cerny:
I have rebased and changed it so that it's now on the rhel8-branch.

But, what do you think about the content of the PR? It's suggesting now that the former method is no longer needed, but I'm not sure whether we can use the reanaconda with kickstart installation.

matejak:
> But, what do you think about the content of the PR? It's suggesting now that the former method is no longer needed, but I'm not sure whether we can use the reanaconda with kickstart installation.

Well, the old method is still documented, so this PR doesn't introduce regressions, and it introduces improvements, so it is a no-brainer. You are right that it is unknown how to combine kickstarts with the reanaconda workflow to us at the moment.

I humbly summon mighty @t184256, the author of reanaconda - we absolutely love the tool wrt testing the GUI installation, is it possible to test different kickstart files as well?

t184256:
Hi, glad to see reanaconda is being used! It was a bit of a shot in the dark because I don't know what the actual users want.

What do you mean by "test different kickstart files", `prime` once and supply different kickstart files in different `update` invocations? I've just checked and ks is fetched after `updates.img` is, so it sounds doable, please file an issue and assign it to me. We'll just need to point `inst.ks` to reanaconda's HTTP server and make it supply different contents.

BTW, if it turns out you have workflows that could benefit from VM snapshotting that won't fit `reanaconda`, I'm open to accomodating them into a much more flexible [fingertip](https://github.com/t184256/fingertip), from which `reanaconda` has descended spiritually.

matejak:
> What do you mean by "test different kickstart files", prime once and supply different kickstart files in different update invocations?

Exactly, that's basically the last feature regarding development support that comes to my mind. Thanks for your interest, will fill an issue!

docs/manual/developer_guide.adoc
```diff
@@ -51,8 +50,66 @@ or use the `lsinitrd` command, which is part of the `dracut` package on RHEL and
...
 
-=== Serve image using HTTP server
+The `reanaconda` script prepares and starts a VM with the update image.
+You can get it from the repository at 
+https://github.com/rhinstaller/devel-tools/tree/master/reanaconda
+
+Prepare the VM:
+
+----
+./reanaconda.py prime --sensible --tree http://ftp.fi.muni.cz/pub/linux/fedora/linux/releases/34/Everything/x86_64/os
```
matejak, line 62:
```suggestion
./reanaconda.py prime --sensible --tree http://ftp.fi.muni.cz/pub/linux/fedora/linux/releases/34/Everything/x86_64/os --variable-kickstart
```

matejak:
Great addition, we should also document the kickstart functionality, but for that, we should provide a base kickstart in the repository.

PR #145, opened 2021-05-26
Raise error when file not on the server or there is another error

matejak:
Previously, the error would be raised when the other code would try to work with the retrieved file that in fact doesn't exist. Now, the error is identified on time, and it can be communicated properly as the installation feedback.

### DIFF
--- a/org_fedora_oscap/data_fetch.py
+++ b/org_fedora_oscap/data_fetch.py
@@ -10,7 +10,11 @@ import os.path
 import pycurl
 
 from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core import constants
+from pyanaconda.threading import threadMgr, AnacondaThread
+from pyanaconda.modules.common.constants.services import NETWORK
 
+from org_fedora_oscap import common
 from org_fedora_oscap.common import _
 from org_fedora_oscap import utils
 
@@ -38,7 +42,7 @@ FILE_URL_RE_STR = r"(file)://(.*)"
 FILE_URL_RE = re.compile(FILE_URL_RE_STR)
 
 
-class DataFetchError(Exception):
+class DataFetchError(common.OSCAPaddonError):
     """Parent class for the exception classes defined in this module."""
 
     pass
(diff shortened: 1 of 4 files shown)


tests/test_data_fetch.py
```diff
@@ -40,6 +42,15 @@ def test_file_retreival():
     assert filecmp.cmp(relative_filename_to_test, temp_filename)
 
 
+def test_file_absent():
+    relative_filename_to_test = "i_am_not_here.file"
+
+    with serve_directory_in_separate_process(PORT):
+        with pytest.raises(data_fetch.FetchError) as exc:
```
jan-cerny, line 49: please check if the exception contain the expected error message text

org_fedora_oscap/data_fetch.py
```diff
@@ -196,3 +196,11 @@ def _curl_fetch(url, out_file, ca_certs=None):
         else:
             msg = "Failed to fetch data: %s" % err
             raise FetchError(msg)
+
+    if protocol in ("http", "https"):
+        return_code = curl.getinfo(pycurl.HTTP_CODE)
+        if 400 <= return_code < 600:
+            msg = (
+                "Failed to fetch data - the request returned HTTP error code {return_code}"
```
jan-cerny, line 204: Missing `f`

jan-cerny (commented):
@matejak  How can I reproduce the problem?

org_fedora_oscap/data_fetch.py
jan-cerny: Sorry, this is .format, disregard my comment, please

matejak: It's probably better to move to f, RHEL8 allows that and it removes that annoying boilerplate.

matejak:
One can reproduce the problem by trying to download files that don't exist. Curl will place a placeholder XML at the download location, and the code will fail later, because the file is not a valid SCAP content. This allows to produce a more helpful error message.

jan-cerny:
> One can reproduce the problem by trying to download files that don't exist. Curl will place a placeholder XML at the download location, and the code will fail later, because the file is not a valid SCAP content. This allows to produce a more helpful error message.

I suppose I won't be able to reproduce it in GUI, but using a kickstart and text mode installation, right?

org_fedora_oscap/data_fetch.py
```diff
@@ -10,7 +10,12 @@
 import pycurl
 
 from pyanaconda.core.configuration.anaconda import conf
+from pyanaconda.core import constants
+from pyanaconda.threading import threadMgr, AnacondaThread
+from pyanaconda.modules.common.constants.services import NETWORK
 
+from org_fedora_oscap import common
+from org_fedora_oscap.common import _
```
jan-cerny, line 18: I'm getting an error that it cant import name '_'.

matejak: I have tested the code, and it doesn't manifest this issue - and `_` is even defined in `common`. How do you bump into this error?

jan-cerny: I run a kickstart installation with a broken URL in the kickstart. But I can't reproduce this problem anymore. It's possible that I did a mistake on Friday. Please disregard.

jan-cerny:
I have tested with this kickstart:

```
# this is a simple kickstart file for testing OSCAP addon's features

lang en_US.UTF-8
timezone Europe/Prague
rootpw aaaaa
bootloader --location=mbr
clearpart --initlabel --all
autopart --type=plain
cmdline
#graphical

url --url=http://download.eng.bos.redhat.com/rhel-8/rel-eng/RHEL-8/latest-RHEL-8/compose/BaseOS/x86_64/os/
repo --name appstream --baseurl=http://download.eng.bos.redhat.com/rhel-8/rel-eng/RHEL-8/latest-RHEL-8/compose/AppStream/x86_64/os/

%packages
vim
%end

%addon org_fedora_oscap
    content-type = datastream
    content-url = http://10.0.2.2:8000/dddddddddd
    profile = xccdf_org.ssgproject.content_profile_standardXXXX
%end
```

Previously, it failed on oscap. Now, it gives a nice error message. Works great!

PR #146, opened 2021-05-26
Introduce various small fixes and refactoring

matejak:
Recommended review workflow is to go by commits - each of them addresses a specific improvements, but it would be too small for a standalone PR.

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -306,8 +306,9 @@ def extract_data(archive, out_dir, ensure_has_files=None):
         # ZIP file
         try:
             zfile = zipfile.ZipFile(archive, "r")
-        except zipfile.BadZipfile as err:
-            raise ExtractionError(str(err))
+        except Exception as exc:
+            msg = "Error exctracting archive as a zipfile: {exc}".format(exc=str(exc))
+            raise ExtractionError(msg)
 
         # generator for the paths of the files found in the archive (dirs end
         # with "/")
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -63,13 +63,20 @@ INSTALLATION_CONTENT_DIR = "/tmp/openscap_data/"
 TARGET_CONTENT_DIR = "/root/openscap_data/"
 
 SSG_DIR = "/usr/share/xml/scap/ssg/content/"
-SSG_CONTENT = "ssg-rhel7-ds.xml"
-if constants.shortProductName != 'anaconda':
-    if constants.shortProductName == 'fedora':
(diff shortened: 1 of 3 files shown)


org_fedora_oscap/common.py
```diff
@@ -341,8 +341,9 @@ def extract_data(archive, out_dir, ensure_has_files=None):
         # ZIP file
         try:
             zfile = zipfile.ZipFile(archive, "r")
-        except zipfile.BadZipfile as err:
-            raise ExtractionError(str(err))
+        except Exception as exc:
+            msg = "Error exctracting archive as a zipfile: {exc}".format(exc=str(exc))
```
jan-cerny, line 359: extracting

org_fedora_oscap/common.py
```diff
@@ -66,13 +66,20 @@ def N_(string): return string
...
 SSG_DIR = "/usr/share/xml/scap/ssg/content/"
-SSG_CONTENT = "ssg-rhel7-ds.xml"
-if constants.shortProductName != 'anaconda':
-    if constants.shortProductName == 'fedora':
-        SSG_CONTENT  = "ssg-fedora-ds.xml"
-    else:
-        SSG_CONTENT = "ssg-%s%s-ds.xml" % (constants.shortProductName,
-                                            constants.productVersion.strip(".")[0])
+
+# Enable patches that set the content name at package-time
```
jan-cerny, line 70: What is  "low-risk monkey-patching"? How can I use it? How do I set the content name at package-time?

matejak: My thought was that the packaging step could invoke `sed` to set e.g. `DEFAULT_SSG_CONTENT_NAME = 'ssg-rhv4-ds.xml'`, in other words, we one could set the content filename manually especially in cases when we know autodetection won't work.

jan-cerny:
@matejak Please resolve the conflict.

org_fedora_oscap/content_handling.py
```diff
@@ -149,15 +152,19 @@ def explore_content_files(fpaths):
     """
 
     def get_doc_type(file_path):
+        content_type = "unknown"
         try:
             for line in execReadlines("oscap", ["info", file_path]):
                 if line.startswith("Document type:"):
                     _prefix, _sep, type_info = line.partition(":")
-                    return type_info.strip()
+                    content_type = type_info.strip()
```
jan-cerny, line 160: Shouldn't a `break` follow this line if there was a `return` before?

jan-cerny:
Thanks, it's a great improvement, I can see the new messages in `/tmp/anaconda.log`.

PR #147, opened 2021-06-04
Replace DataStreamHandler and BenchmarkHandler

jan-cerny:
The new class SCAPContentHandler can handle SCAP source data
streams, XCCDF files and tailoring files using common code.
The classes DataStreamHandler and BenchmarkHandler are removed.
The purpose of the new class is the same as of the removed
classes. The purpose is to get a list of profiles (their IDs
and descriptions) to be able to show it in the GUI. It is
used only in the GUI.


jan-cerny:
I'm thinking how to test it and I have already found multiple combinations that we need to test:

1. GUI installation without kickstart
a. scap-security-guide
b. data stream
c. RPM cotaining data stream
d. rpm containing XCCDF
e. archive containing data stream
f. archive containing XCCDF
2. GUI installation with kickstart
a. scap-security-guide
b. data stream
c. RPM cotaining data stream
d. RPM cotaining data stream + tailoring
e. rpm containing XCCDF
f. rpm containing XCCDF + tailoring
g. archive containing data stream
h. archive containing data stream + tailoring
i. archive containing XCCDF
j. archive containing XCCDF + tailoring

These classes are used only in GUI so I think we don't need to test the text mode.

org_fedora_oscap/scap_content_handler.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def _parse_profiles_from_xccdf(self, benchmark):
+        if benchmark is None:
+            return []
+        if benchmark.tag.startswith(f"{{{ns['xccdf-1.1']}}}"):
+            xccdf_ns_prefix = "xccdf-1.1"
+        elif benchmark.tag.startswith(f"{{{ns['xccdf-1.2']}}}"):
+            xccdf_ns_prefix = "xccdf-1.2"
+        else:
+            raise SCAPContentHandlerError("Unsupported XML namespace")
```
matejak, line 108: Would be nice to indicate what namespace is unsupported.

org_fedora_oscap/scap_content_handler.py
```diff
@@ -0,0 +1,205 @@
...
+        :param checklist_id: value of ds:component-ref/@id pointing to
+            an xccdf:Benchmark
+        :type checklist_id: str
+        :return: None
+
+        """
+        self._data_stream_id = data_stream_id
+        self._checklist_id = checklist_id
+
+    def get_profiles(self):
```
matejak, line 146: Those nested ifs are confusing. I propose to restructure it in the following way:

Firs identify all pathological cases, throw all the easy errors - bad XCCDF, standalone tailoring or unknown content.
Then, there is the trivial benchmark extraction from XCCDF, and multi-step benchmark extraction from a datastream, that I propose to extract to another method.

org_fedora_oscap/scap_content_handler.py
```diff
@@ -0,0 +1,205 @@
...
+        :param file_path: path to an SCAP file (only SCAP source data streams,
+        XCCDF files and tailoring files are supported)
+        :type file_path: str
+        :param tailoring_file_path: path to the tailoring file, can be None if no tailoring exists
+        :type tailoring_file_path: str
+        """
+        self.file_path = file_path
+        tree = ET.parse(file_path)
+        self.root = tree.getroot()
+        if tailoring_file_path is not None:
```
matejak, line 58: You can reduce to a Pythonic `if not tailoring_file_path`, which would handle empty strings as well, and you could therefore streamline some of the higher-level code.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -427,35 +426,38 @@ def _init_after_data_fetch(self, wait_for):
...
             raise common.OSCAPaddonError("Unsupported content type")
 
         try:
-            self._content_handler = self._content_handling_cls(self._addon_data.preinst_content_path,
-                                                               self._addon_data.preinst_tailoring_path)
-        except content_handling.ContentHandlingError:
+            if self._addon_data.preinst_tailoring_path:
+                tailoring_path = self._addon_data.preinst_tailoring_path
+            else:
+                tailoring_path = None
```
matejak, line 445: This can be streamlined - see another comment below.

tests/test_scap_content_handler.py
```diff
@@ -0,0 +1,193 @@
...
+    checklists = ch.get_data_streams_checklists()
+    assert checklists is None
+    profiles = ch.get_profiles()
+    assert len(profiles) == 1
+    pinfo1 = ProfileInfo(
+        id="default",
+        title="Default",
+        description="The implicit XCCDF profile. Usually, the default profile "
+        "contains no rules.")
+    assert pinfo1 in profiles
```
matejak, line 193: Please add a newline, text files that don't end by a newline are considered dubious.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -581,18 +583,13 @@ def _update_profiles_store(self):
...
 
         self._profiles_store.clear()
-
-        if self._using_ds:
-            profiles = self._content_handler.get_profiles(self._current_ds_id,
-                                                          self._current_xccdf_id)
-        else:
-            # pylint: disable-msg=E1103
-            profiles = self._content_handler.profiles
+        profiles = self._content_handler.get_profiles()
```
matejak, line 592: This can lead to exceptions, so what about dispatching a logging message that something went wrong with the update of profiles, and pass the exception on?
The previous behavior was equally bad about this, and any exception would probably kill the addon, so I wouldn't concentrate on this aspect in this PR.

matejak (changes requested):
I have pointed out small improvements, and an issue when the profile update goes wrong.

matejak (approved):

matejak:
The handler works as expected, and it is also much faster than the previous implementation. Great work, thanks!

PR #148, opened 2021-06-04
Refactor handling of content

matejak:
This PR aims to process content in a way that is backend-agnostic, i.e. it can work for the automated kickstart installations as well as interactive ones. It is separated into multiple commits - most of them are small, but the main one is quite extensive.

The fetch is performed in a two-step manner that was present in the GUI code, when it is started in a separate thread, and then the processing code then waits for the fetching thread - this may be an overkill for a plain kickstart installation, but it doesn't cause any problems.
Interactions with the backend are performed by callbacks - at this time it is a possibility to report status, definition of what to do after fetching stops, and finally definition of how to handle errors by means of a function that accepts an exception instance.
The unified pipeline includes content fetching, fingerprint verification, file extraction and file identification.

A class containing a content set has been introduced - it contains a list of discovered files (fetched, present on the system, or extracted from an archive) that have been identified by the `oscap` scanner. That class allows to work with the content more easily.


org_fedora_oscap/model.py
```diff
@@ -0,0 +1,281 @@
...
+    def _verify_fingerprint(self, dest_filename, fingerprint=""):
+        if not fingerprint:
+            return
+
+        hash_obj = utils.get_hashing_algorithm(fingerprint)
+        digest = utils.get_file_fingerprint(dest_filename,
+                                            hash_obj)
+        if digest != fingerprint:
+            log.error(
+                "File {dest_filename} failed integrity check - assumed a "
```
jan-cerny, line 134: Missing f.

org_fedora_oscap/model.py
```diff
@@ -0,0 +1,281 @@
...
+        for f, l in labelled_files.items():
+            content.add_file(f, l)
+
+        if fingerprint:
+            content.record_verification(dest_filename)
+
+        return content
+
+
+class ObtainedContent:
```
jan-cerny, line 190: I feel that it somehow overlaps with OSCAPData class (forms ks/oscap.py).

tests/test_data_fetch.py
```diff
@@ -59,3 +59,11 @@ def test_supported_url():
 
 def test_unsupported_url():
     assert not data_fetch.can_fetch_from("aaaaa")
+
+
+def test_fetch_local(tmp_path):
+    source_path = pathlib.Path(__file__).absolute()
+    dest_path = tmp_path / "dest"
+    data_fetch.fetch_data("file://" + str(source_path), dest_path)
```
jan-cerny, line 67: So if I am an user when I run the installation where do I put the files or how do I pass files to Anaconda?

org_fedora_oscap/model.py
```diff
@@ -0,0 +1,281 @@
...
+log = logging.getLogger("anaconda")
+
+
+def is_network(scheme):
+    return any(
+        scheme.startswith(net_prefix)
+        for net_prefix in data_fetch.NET_URL_PREFIXES)
+
+
+class Model:
```
jan-cerny, line 24: Can we think of a better name?

org_fedora_oscap/model.py
```diff
@@ -0,0 +1,281 @@
...
+            raise content_handling.ContentHandlingError(msg)
+
+    def select_main_usable_content(self):
+        elligible_main_content = (self._datastream_content(), self._xccdf_content())
+        if not any(elligible_main_content):
+            msg = (
+                "Couldn't find a valid datastream or a valid XCCDF-OVAL file tuple "
+                "among the available content")
+            raise content_handling.ContentHandlingError(msg)
+        if elligible_main_content[0]:
```
jan-cerny, line 267: this perhaps require a comment or better unpacking the tuple on line  into 2 variables that would have good names and changing the condition.

org_fedora_oscap/model.py
```diff
@@ -0,0 +1,281 @@
...
+    def content_uri(self):
+        return self.content_uri_scheme + "://" + self.content_uri_path
+
+    @content_uri.setter
+    def content_uri(self, uri):
+        scheme, path = uri.split("://", 1)
+        self.content_uri_path = path
+        self.content_uri_scheme = scheme
+
+    def fetch_content(self, cert, what_if_fail):
```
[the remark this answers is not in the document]
matejak: That makes sense now, but what if you would like to introduce support for fetching tailoring? Then, it would make sense to keep those separately.

org_fedora_oscap/model.py
matejak: This one aims to assist the gathered files discovery - files can be downloaded directly, or they can be extracted. This class enables a user to quickly understand what is available, and whether the current set of contents is usable for further processing.

tests/test_data_fetch.py
matejak: You can use this only if those files are already there. I can imagine usage when the content filename is not guessed properly (e.g. RHEL8 vs RHV), or when the installation image contains the content as a specific location.

org_fedora_oscap/model.py
matejak: or one can avoid the tuple completely.

jan-cerny (commented):
It mostly works well.

See the comments below.

Please check the unit tests, they don't pass for me.

org_fedora_oscap/ks/oscap.py
```diff
@@ -408,86 +427,41 @@ def setup(self, storage, ksdata, payload):
...
-                # run mode and let things go on
-                self.dry_run = True
-                return
-            else:
-                # Let's sleep forever to prevent any further actions and wait
-                # for the main thread to quit the process.
-                progressQ.send_quit(1)
-                while True:
-                    time.sleep(100000)
+            msg = ["Wrong configuration detected!"]
```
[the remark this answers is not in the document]
matejak: Another weird thing is the absence of translations.

tests/test_content_handling.py
```diff
@@ -0,0 +1,101 @@
...
+PROFILE3_ID = "xccdf_com.example_profile_my_profile3"
+
+
+@pytest.fixture()
+def ds_handler():
+    return ch.DataStreamHandler(DS_FILEPATH)
+
+
+def test_init_invalid_file_path():
+    with pytest.raises(ch.DataStreamHandlingError) as excinfo:
```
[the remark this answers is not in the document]
matejak: Ouch.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -388,62 +398,33 @@ def _init_after_data_fetch(self, wait_for):
...
-                                               files.tailoring)
-        elif self._addon_data.content_type not in (
-                "datastream", "scap-security-guide"):
-            raise common.OSCAPaddonError("Unsupported content type")
+            return
 
         try:
+            if actually_fetched_content:
+                self.content_bringer.use_downloaded_content(content)
+            log.info(f"Opening SCAP content at {self._addon_data.preinst_content_path} with tailoring {self._addon_data.preinst_tailoring_path}")
```
jan-cerny, line 422: I'm sorry but now I realized that self._addon_data.preinst_tailoring_path isn't set when tailoring isn't present so you would get: "Opening SCAP content at /tmp/bzfn with tailoring" in that situation which would be confusing for people reading the log.

jan-cerny (commented):
I have tried multiple different content from the testing_files directory and it seems to work. Great improvement.

I have only one small nitpick.

org_fedora_oscap/gui/spokes/oscap.py
matejak: Fixed by a rebase.

jan-cerny (approved):
amazing

PR #149, opened 2021-06-08
Add testing files

jan-cerny:
This adds more files that can be used in unit tests or for manual
testing. The commit adds also a README file where you can read
their descriptions.


matejak:
I probably wouldn't be so verbose in the README and I would aim to express everything in the filename, but that's for future generations.

PR #152, opened 2021-06-10
Import mock from unittest, where it has been relocated

matejak:
The separate Python mock package has been deprecated as a result of the mock inclusion into the standard library.


jan-cerny:
@matejak Please don't insert the dots where they shouldn't be because I'm tired from removing them.

jan-cerny (approved):

PR #153, opened 2021-06-11
GH Actions: Init with unittest

evgenyz:


matejak:
This looks good, but we will need to be able to adjust a branch-Fedora version mapping somehow. For example, the `master` branch corresponds to rawhide, and `rhel9` to 34 for the time being.

evgenyz:
> This looks good, but we will need to be able to adjust a branch-Fedora version mapping somehow. For example, the `master` branch corresponds to rawhide, and `rhel9` to 34 for the time being.

What about rhel8? F33?

evgenyz:
The best I could come up with ATM.

matejak:
So rawhide is not a good target, as even casual packages don't have to be installable there every day, and I have estimated Fedora 32 to be a good choice for the rhel8-branch.
I am merging this, the rest of settings looks reasonable, thanks for the investigation and effort!

matejak (approved):

PR #155, opened 2021-06-14
GH Actions: Single job for rhel8 branch

evgenyz:


matejak (approved):
That's wonderful - tests pass, and the configuration is local.

PR #156, opened 2021-06-14
GH Actions: Single action for rhel9 branch

evgenyz:


matejak (approved):
Tests are executed correctly. Failures are caused by misalignment between tests and the Anaconda runtime.

PR #157, opened 2021-06-14
GH Actions: Single action for 'master' branch

evgenyz:


matejak (approved):
Test configuration is correct. Tests are not expected to pass due to infrastructure issues when an early development version of Fedora is used.

PR #158, opened 2021-06-16
Merge RHEL8 changes to the RHEL9 branch

matejak:


jan-cerny (approved):
I have spotted a problem with using zip files on a RHEL 9 machine, however I think that the problem isn't related to this PR, because I can reproduce it in REPL.  The error is `unknown encoding: cp437` in  a `zipfile.ZipFile(...) call`.

PR #159, opened 2021-06-24
Polish some aspects of the addon

matejak:
- Try to make it not only a "fedora" addon, but also a proper "rhel" addon.
- Take away the fingerprint verification task - that one is inseparable from the content fetch.
- Reuse the `PolicyData` structure in the kickstart.
- Change the question in case something goes wrong during the installation.
- Properly convert messages to strings when printing them.
- Handle the case when the content has been already fetched during the first phase - in that case, just be sure to come up with an error if the content is defective.

### DIFF
--- a/org_fedora_oscap/service/installation.py
+++ b/org_fedora_oscap/service/installation.py
@@ -77,7 +77,7 @@ class PrepareValidContent(Task):
         """Run the task."""
         # Is the content available?
         fetching_thread_name = None
-        if not os.path.exists(self._content_path) and not os.path.exists(self._file_path):
+        if not os.path.exists(self._content_path):
             # content not available/fetched yet
             fetching_thread_name = self.content_bringer.fetch_content(
                 _handle_error, self._policy_data.certificates)
@@ -91,7 +91,15 @@ class PrepareValidContent(Task):
             lambda msg: log.info(msg), content_dest, _handle_error)
 
         if not content:
-            return
+            # this shouldn't happen because error handling is supposed to
+            # terminate the addon before finish_content_fetch returns
+            _handle_error(Exception())
+
+        remote_content_was_present = (
+            not fetching_thread_name
+            and self._policy_data.content_type != "scap-security-guide")
+        if remote_content_was_present:
+            content.add_file(self._content_path)
 
         try:
             # just check that preferred content exists
--- a/org_fedora_oscap/service/installation.py
+++ b/org_fedora_oscap/service/installation.py
@@ -54,7 +54,6 @@ def _handle_error(exception):
 
 def terminate(message):
     message += "\n" + _("The installation should be aborted.")
-    message += " " + _("Do you wish to continue anyway?")
     raise NonCriticalInstallationError(message)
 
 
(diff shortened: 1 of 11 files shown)


org_fedora_oscap/service/installation.py
```diff
@@ -25,92 +25,76 @@
...
 
-        try:
-            data_fetch.fetch_data(
-                self._policy_data.content_url,
-                self._file_path,
-                self._policy_data.certificates
-            )
+def terminate(message):
+    message += "\n" + _("The installation should be aborted.")
+    message += " " + _("Do you wish to continue anyway?")
```
poncovka, line 56: Just a note, our handler of the `NonCriticalInstallationError` exception produces a dialog with the "Would you like to ignore this and continue with installation?" question (see https://github.com/rhinstaller/anaconda/pull/3147#issuecomment-773557362). It might be confusing to ask two questions.

matejak: Is there a difference between RHEL8 Anaconda and the RHEL9 one? I recall that I have seen an error message without the question probably in the RHEL8 GUI, and there were yes/no buttons at the bottom of the message window, which looked weird without the actual question.
I am not 100% sure how to reproduce this, but I can try. Do you have any idea what was it about?

poncovka: In RHEL 8, the addon had to handle a non critical error and show a dialog. In RHEL 9, the addon is not able to handle a non critical error, because the installation tasks are run in a DBus module, so the installer (the process with UI) catches and handles the `NonCriticalInstallationError` error during the installation.

You can try to modify the installation task to make sure that it will raise the error.

org_fedora_oscap/service/kickstart.py
```diff
@@ -249,5 +236,7 @@ class OSCAPKickstartSpecification(KickstartSpecification):
     """The kickstart specification of the OSCAP service."""
 
     addons = {
-        "org_fedora_oscap": OSCAPKickstartData
+        "org_fedora_oscap": OSCAPKickstartData,
+        "org_centos_oscap": OSCAPKickstartData,
+        "com_redhat_oscap": OSCAPKickstartData,
```
poncovka, line 241: These variants can be tested in unit tests. There is also support for checking errors and warnings.

jan-cerny (changes requested):
Making a commit that fixes multiple unrelated things such as cd691cc is considered a bad practice. 

Also, do not end the subject line with a period. Use the imperative mood in the subject line of your commit messages. You can read https://chris.beams.io/posts/git-commit/ which is a nice article on git commits.

When I loaded testing_data/testing_ds.xml, selected "My testing profile" and clicked on the "Select profile" button I get a dialog "An unknown error has occured" window which contains an exception
Screenshot from 2021-06-30 15-44-28

I haven't checked whether it's caused by this PR or whether it was there before. Can you check it?

matejak:
Thanks for the review, you are right about good practices - thanks for the read!

Regarding password problem, I have decided to simply block it, as it is an Anaconda API issue, and the respective content is only present in the testing datastream - it is not shipped anywhere in the scap-security-guide, and as the Anaconda remediation format is not standardised, it won't appear anywhere else.
Moreover the feature works in a strange way - much more than just some password length is enforced in the case of the testing DS.

poncovka:
Please, see https://github.com/OpenSCAP/oscap-anaconda-addon/pull/103/files#r662201184.

jan-cerny (approved):

PR #160, opened 2021-06-25
Add a TUI to the addon

matejak:
The TUI interface is one of the two interactive interfaces an Anaconda addon may implement.

### DIFF
--- a/org_fedora_oscap/tui/spokes/oscap.py
+++ b/org_fedora_oscap/tui/spokes/oscap.py
@@ -1,0 +1,217 @@
+#
+# Copyright (C) 2013  Red Hat, Inc.
+#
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/tui/spokes/oscap.py
```diff
@@ -0,0 +1,217 @@
...
+
+# import gettext
+# _ = lambda x: gettext.ldgettext("hello-world-anaconda-plugin", x)
+
+# will never be translated
+_ = lambda x: x
+N_ = lambda x: x
+
+
+class OSCAPTuiSpoke(FirstbootSpokeMixIn, NormalTUISpoke):
```
poncovka, line 58: Are you sure you want to run this spoke in the initial setup on first boot?

PR #162, opened 2021-06-30
Fix multiple small RHEL8 issues

matejak:
This is a sister PR of #159 that fixes fatal omissions not discovered during testing so far.


jan-cerny (approved):
LGTM

PR #163, opened 2021-07-07
Post-release update

matejak:
This PR fixes issues found right after performing the upstream release.

- Bump the release number
- Update the translation template and fix the branch for translations.
- Fix `make dist`


jan-cerny:
@matejak The commit 0207863 has deleted the spec file, but unfortunately, the commit message doesn't explain why it is removed.

matejak:
So now the commit message explains that:

> The project requires a RHEL spec file to function properly, which is is not an upstream type of content.

jan-cerny (approved):

PR #167, opened 2021-08-03
Refactor content identification

matejak:
Don't use the multiprocessing pool - it sometimes creates probems during its initialization: https://bugzilla.redhat.com/show_bug.cgi?id=1989441

### DIFF
--- a/org_fedora_oscap/content_handling.py
+++ b/org_fedora_oscap/content_handling.py
@@ -111,9 +111,8 @@ ContentFiles = namedtuple("ContentFiles", ["xccdf", "cpe", "tailoring"])
 
 
 def identify_files(fpaths):
-    with multiprocessing.Pool(os.cpu_count()) as p:
-        labels = p.map(get_doc_type, fpaths)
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/content_handling.py
```diff
@@ -131,7 +130,9 @@ def get_doc_type(file_path):
     except UnicodeDecodeError:
         # 'oscap info' supplied weird output, which happens when it tries
         # to explain why it can't examine e.g. a JPG.
-        return None
+        pass
+    except Exception as e:
+        log.warning(f"OSCAP addon: Unexpected error when looking at {file_path}: str(e)")
```
jan-cerny, line 135:
```suggestion
        log.warning(f"OSCAP addon: Unexpected error when looking at {file_path}: {str(e)}")
```

jan-cerny (changes requested):

matejak:
I have fixed that by a rebase, so we don't introduce micro commits.

jan-cerny (approved):

PR #168, opened 2021-08-09
Fix archive handling in GUI installs

matejak:
GUI downloads an archive, so the ensuing installation doesn't have to.
However, the installation has to be able to discover files recovered from the archive.
The fix makes sure that files are discovered also in subdirectories.


jan-cerny (approved):
For me it fixes the problem with an error message after clicking on Begin installation when using RPM content.

PR #169, opened 2021-08-18
Don't show the OSCAP spoke if the OSCAP DBus module is disabled

poncovka:
Add-ons can be disabled in the Anaconda configuration files. Without the fix,
the OSCAP DBus module is started on demand by the OSCAP spoke even though it
shouldn't be activated. In the future, it will result in a failure of the
installer.

Related: rhbz#1994003


matejak:
Please rebase against the current master.

poncovka:
Rebased.

matejak:
LGTM, this PR obviously doesn't break backwards compatibility, and I am sure that you know what you are doing :-)
Could you be so kind and also create a RHEL9 rebase of the PR?

matejak (approved):

PR #174, opened 2021-09-27
Allow to create an update image in a container

poncovka:
Run `make container-update-image` to create an update image of the OSCAP
addon and its main dependencies in a container.

Download and install the `xmlsec1-openssl` dependency for the upstream testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,10 @@ container-test:
 	podman build --tag $(CONTAINER_NAME) --file tests/Dockerfile
 	podman run --volume .:/oscap-anaconda-addon:Z $(CONTAINER_NAME) make test
 
+container-update-image:
+	podman build --tag $(CONTAINER_NAME) --file tests/Dockerfile
+	podman run --volume .:/oscap-anaconda-addon:Z $(CONTAINER_NAME) ./create_update_image.sh -r / download_rpms
+
 test: unittest runpylint
 
 runpylint:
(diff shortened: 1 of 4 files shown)


poncovka:
Rebased.

Makefile
```diff
@@ -135,6 +135,10 @@ container-test:
 	podman build --tag $(CONTAINER_NAME) --file tests/Dockerfile
 	podman run --volume .:/oscap-anaconda-addon:Z $(CONTAINER_NAME) make test
 
+container-update-image:
+	podman build --tag $(CONTAINER_NAME) --file tests/Dockerfile
+	podman run --volume .:/oscap-anaconda-addon:Z $(CONTAINER_NAME) ./create_update_image.sh download_rpms
```
matejak, line 140: The problem here is that the script doesn't know what version of to download - it should be given the `-r` argument, as otherwise it tries to get the F28 packages, as it is a default from old times.
Any idea how to determine the latest Fedora release programmatically, so we could determine the version number of rawhide? A very dirty hack would be to deduce that from the calendar.

poncovka: I have changed the default releasever of the script to rawhide and set the releasever for the container to the host releasever. However, I did some testing and DNF will use whatever releasever+ packages anyway if it doesn't have access to the right repositories. There were no differences between running  `make container-update-image` with `28` and `rawhide` releasevers.

matejak: OK, I didn't know that. But it's great that besides numerical versions, the `rawhide` `releasever` is recognized by `dnf` as well. It totally makes sense to have that one as the default.

matejak (approved):
LGTM, thank you for the PR!

PR #176, opened 2021-10-15
Represent unselected profile by an empty string

matejak:
`None` can't be passed via the DBUS interface.

There are no other obvious leftover `None`s in the GUI code - the data sent over DBUS is defined in the `apply` method of the GUI spoke.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2014108


jan-cerny:
On the latest RHEL9, it fixes the traceback mentioned in  https://bugzilla.redhat.com/show_bug.cgi?id=2014108 for me.

jan-cerny (approved):

PR #177, opened 2021-10-26
Fix bad destination for the parsed content fingerprint

matejak:
Using the old ways, the fingerprint was not propagated to the post-install kickstart.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1993065


jan-cerny:
I have checked  at at the end of the installation the fingeprint appears in /mnt/sysroot/root/anaconda-ks.cf. Moreover, the fingeprint comparison now actually started to work, it successfully prevents loading the content that doesn't match the fingerprint!

jan-cerny (approved):

PR #178, opened 2021-11-08
Merge rhel9 branch into master

matejak:
The RHEL9 got some fixes that we want in master as well.

In another PR, we will take goodies that got into the master, trying to propagate them to RHEL9.

PR #179, opened 2021-11-19
Use the PackagesSelection property

poncovka:
The `Packages` property no longer exists. It was replaced with `PackagesSelection` and
`PackagesConfiguration` properties. This bug was introduced in the commit 2014f1e.


matejak:
Commits have been moved to #178, so I will close this, as that PR is likely to get merged.

PR #181, opened 2021-12-01
Fix and enable tests for the upstream container

poncovka:
* Install new dependencies in the testing container.
* Don't skip the `test_scap_security_guide` test.
* Don't skip the `test_evaluate_rules_task` test.


matejak:
LGTM, thanks!

matejak (approved):

PR #183, opened 2022-01-03
Rhel9 - create update image in a container

matejak:
Port of #174 to RHEL9


jan-cerny (approved):
I tried to execute it. It has built the update.img successfully.

PR #184, opened 2022-01-03
Improve UX when there are weird errors when attempting remediation

matejak:
This PR fixes these related issues:

- Obvious scanner errors are communicated clearly before any serious work involving the scanner is performed.
- Errors result in installer questions rather than in tracebacks.
- The installer feedback has been adjusted to fit the installation mode, so it doesn't ask questions that can't be answered.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2007981


jan-cerny:
I have performed an installation of 8.5 in a Reanaconda environment proving only the BaseOS http repo as an installation source. With this patch, the traceback and the error pop up window no longer appears.

jan-cerny (approved):

PR #191, opened 2022-01-19
Document testing of OAA with KS

jan-cerny:
- describes method using virt-install
- adds an example kickstart

### DIFF
--- a/docs/manual/developer_guide.adoc
+++ b/docs/manual/developer_guide.adoc
@@ -155,9 +155,39 @@ Watch the console, as the VM is supposed to download the update image, and the P
 
 200 is the OK request status.
 
+==== Installing a VM using update image and kickstart
 
+You can also use the `virt-install` command, which is useful when you want to test kickstart installation.
 
+Some kickstarts (`.cfg` files) can be found in the `testing_files` directory.
+You will also need installation URL and the update image described above.
 
+For example:
+
+----
+virt-install \
+--connect qemu:///system \
+--name oaa_test \
+--memory 2048 --vcpus 2 --disk size=8 \
+--os-variant rhel8.5 \
+--location INSTALLATION_URL \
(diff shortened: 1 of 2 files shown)


docs/manual/developer_guide.adoc
```diff
@@ -155,9 +155,39 @@ Watch the console, as the VM is supposed to download the update image, and the P
...
+
+----
+virt-install \
+--connect qemu:///system \
+--name oaa_test \
+--memory 2048 --vcpus 2 --disk size=8 \
+--os-variant rhel8.5 \
+--location INSTALLATION_URL \
+-x inst.updates=http://gateway:8000/update.img  \
+--initrd-inject /path/to/ks.cfg \
```
matejak, line 175: Would it be possible to inject the update image as well by supplying it multiple times? The network communication with the host is more fragile, so it would be nice to be able to get rid of it.
OTOH, it is possible to supply `inst.ks=http://gateway:8000/ks.cfg` in the same way as one specifies the update image. So if you can't inject everything into the initrd, I would suggest to get rid of the injection and use the network for both the ks and the update image.

jan-cerny: Great idea! I tried to inject update image but it didn't work. Also, it isn't described in https://fedoraproject.org/wiki/Anaconda/Updates. On the other hand, serving kickstart over web server works and is documented. So I changed this to unify and use web server for both updates image and kickstart.

jan-cerny:
@matejak anything else needed?

matejak:
Great, I can confirm that it works, unless there are issues on the local computer when the firewall may be blocking some network connections, but that's unrelated.

So could you please open adjusted PRs for RHEL9 and Fedora as well?

matejak (approved):

matejak:
@jan-cerny Could you please open adjusted PRs for RHEL9 and Fedora as well?

PR #193, opened 2022-03-09
Use inst.ks in the virt-install command

jan-cerny:
Using just "ks" is deprecated.


matejak (approved):
LGTM, thank you for the PR and please also submit one for rawhide.

PR #200, opened 2022-04-20
Add Packit configuration

matusmarhefka:
Build RPM on all supported Fedora versions using Packit service.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,11 +6,11 @@ on:
     branches: [ master ]
 jobs:
   validate-master:
-    name: Unit Tests on Fedora Rawhide (Container)
+    name: Unit Tests on Fedora (Container)
     if: github.ref == 'refs/heads/master' || github.event.pull_request.base.ref == 'master'
     runs-on: ubuntu-latest
     container:
-      image: fedora:35
+      image: fedora:latest
     steps:
       - name: Install Deps
         run: dnf install -y make anaconda python3-pytest python3-pycurl openscap-scanner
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     if: github.ref == 'refs/heads/master' || github.event.pull_request.base.ref == 'master'
     runs-on: ubuntu-latest
     container:
-      image: fedora:35
+      image: fedora:latest
```
matusmarhefka, line 13: @matejak This will pull the latest released Fedora container, not rawhide. Should we rather change this to `fedora:rawhide` image?

Makefile
```diff
@@ -1,6 +1,6 @@
 NAME = oscap-anaconda-addon
 
-VERSION = 0.35.0
```
matejak, line 3: The versioning is tricky.
1.x are RHEL8-compatible versions
2.x are RHEL9-compatible versions, and so on.
In case of Fedora, a scheme 0.x.y is used, where x is the respective Fedora version. OTOH, as the addon is not used on Fedora ATM, the likelihood of making a Fedora upstream release is relatively low.

.github/workflows/build.yaml
matejak: This is probably OK, as it is about running tests, and they are not so tightly coupled with the Anaconda, and other things are more likely to cause problems in Rawhide.

Makefile
matusmarhefka: Ok, will do.

matejak (approved):
LGTM, thank you for the PR!

PR #211, opened 2022-05-23
Remove the oscap-utils dependency

matejak:
oscap-utils package is not related to the firstboot remediation,
everything will continue to work even without it being an installation requirement


jan-cerny:
This is a port of already merged PR https://github.com/OpenSCAP/oscap-anaconda-addon/pull/209 to rhel8-branch.

jan-cerny (approved):

PR #212, opened 2022-06-07
Use Anaconda DBus read-write properties

poncovka:
Anaconda is replacing `Set.*` methods with read-write properties to clean up
the DBus API and simplify unit testing with mocked DBus objects. This support
was introduced in `anaconda-37.8-1`.


poncovka:
This change is safe to release on Rawhide.

matejak (approved):
LGTM, thanks for the update!

PR #219, opened 2022-10-12
Handle archives and preferred content better

matejak:
The addon lost the ability to pick files from supplied archives based on kickstart specifications.
This PR allows this, which allows the addon to

- handle archives with e.g. multiple datastream files, and to
- throw error when an expected file in the archive is missing.

This PR fixes RHEL8 and RHEL9 BZs

- https://bugzilla.redhat.com/show_bug.cgi?id=2129008
- https://bugzilla.redhat.com/show_bug.cgi?id=2129846

PR #220, opened 2022-10-13
Rhel8 robust archive analysis

matejak:
Port of #219 to RHEL8

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2129008


jan-cerny:
We have discovered that if you pass a RPM package that contains multiple SCAP files and tailoring files, the tailoring is picked even if not requested and if there are multiple tailoring files in there it terminates with the same error as in the referenced BZ. This happens with scap-security-guide-0.1.63-4.el8.noarch.rpm. The message in the log is:
Screenshot from 2022-10-13 13-51-43
.

matejak:
The case that you bumped into indicates that accepting tailoring in a "greedy" fashion is problematic. Therefore, I have updated the functionality to use tailoring only in cases when it is expected.

jan-cerny:
Thnaks for the update. From a quick test it seems that it resolved the aforementioned problem with tailoring files. However, we have discovered a different problem on our way. At later phase, the OAA tries to install the content RPM package. However, if this package has a dependency, it fails to install and breaks the OS installation. Even though not every RPM package requires other packages, it might be a frequent problem, because the `scap-security-guide` RPM that we ship has some required dependencies, and people can use the `rpm` option in the kickstart to use a different  version of the `scap-security-guide` that the version present in the installation image they install from. Specifically, in my example with `scap-security-guide` the installation failed because it could not find xml-common, and that was visible in the log tab in the Anaconda tmux session. Please think of possible solution.

matejak:
Our testing scenarios don't use subscriptions, and as a result, downloading of packages doesn't work out of the box. I think that an approach of "try yum, if it doesn't work because of dependencies, use `rpm -i --nodeps`".

jan-cerny:
We have been we discussed this with @matejak  and we mentioned the following items:

- There is possibility that the use case of using the scap-security-guide rpm as rpm in the kickstart will be one of the frequent use-cases of the rpm kickstart option because some people will want to use a more recent scap-security-guide rpm in their installation than the one that is present in the installation image or in the installation web tree, for example when installing a version that has been updated as a part of the z-stream channel. This might be mitigated on the user's side by extracting the content from the RPM and distributing it in a plain file form, but this workaround can add additional friction for the users.
- Other frequent use cases will be using RPMs created by SCAP Workbench, but with these the issue isn't expected to happen. 

There are multiple possible solutions to the issue.

- First is mentioned in the comment above - to try installation with yum and as a fallback use the `rpm -i --nodeps` command.
-  Other possible solution is to treat RPM packages similarly as we handle the ZIP archives, which means to simply extract all the files from the RPM to their target path. This can be modified further and we can extract only the file that is specified or detected as the main content failed and extract it to a per-defined directory instead of extracting it to the directory specified by the RPM package.
- Third solution is to detect the dependencies of the supplied RPM package in the pre-installation phase where the list of packages to be installed is created by Anaconda and add the dependencies to the list so that in the post phase there will be no unexpected missing dependencies.

The solutions have their advantages and disadvantages, the main concern are possible conflicts with the package dependencies or versions of other packages, possibly a conflict with the package requirements with of security policy itself.

jan-cerny (approved):
I have performed an RHEL 8.6 kickstart installation with conent type  = rpm, profile = anssi_bp28_minimal and a using a different (newer) version of scap-security-guide than in the compose. It all went OK, after installation the given RPM has been present on the system in the expected version and the messages in the /var/log/anaconda/anaconda.log were OK. Great job!

matejak:
Thanks for the review!

PR #221, opened 2022-10-17
Fix clicking frenzy crash for RHEL8

matejak:
Port of #218 to the RHEL8 branch.

Also fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2000998

### DIFF
--- a/org_fedora_oscap/gui/spokes/oscap.py
+++ b/org_fedora_oscap/gui/spokes/oscap.py
@@ -363,11 +363,14 @@ class OSCAPSpoke(NormalSpoke):
         else:
             renderer.set_property("stock-id", None)
 
+    def _still_fetching(self):
+        return self._fetching or threadMgr.get('OSCAPguiWaitForDataFetchThread')
+
     def _fetch_data_and_initialize(self):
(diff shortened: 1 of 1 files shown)


org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -149,8 +149,7 @@ def decorated(self, *args, **kwargs):
         self._unitialized_status = None
         self._ready = True
         # pylint: disable-msg=E1101
-        hubQ.send_ready(self.__class__.__name__, True)
-        hubQ.send_message(self.__class__.__name__, self.status)
+        hubQ.send_ready(self.__class__.__name__)
```
jan-cerny: I got a traceback on this line immediately when loading the main Anaconda window, it's a TypeError Excpetion, expecting 2 arguments but 1 given.

jan-cerny:
```suggestion
        hubQ.send_ready(self.__class__.__name__, True)
```

jan-cerny (approved):
I followed the reproducer in the linked BZs and this patch fixed the problem for me! Thx

PR #224, opened 2022-11-10
Rhel9 archive handling

matejak:
#### Description:

- Port of RHEL8's #220 to RHEL9, the same applies, just the addon code and architecture are different.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2129846


jan-cerny (approved):
I have reproduced the RHEL 9 variant of the bug which can be found at https://bugzilla.redhat.com/show_bug.cgi?id=2129846. Then, I created an update image from the code submitted in this PR and I tried a new installation of a RHEL 9.0 virtual machine with this update image. With this PR the bug doesn't occur during the installation.

PR #225, opened 2022-11-15
Fix handling of content paths

### LINKED ISSUE #827: rhel-8 tests massively timeout on OSCAP spoke failure
First seen on Nov 13 2022: https://github.com/rhinstaller/kickstart-tests/actions/runs/3457378548
Screenshot from local reproducer:
Screenshot from 2022-11-14 12-46-45 oscap_fail
Screenshot from 2022-11-14 12-47-09 oscap_fail
[kstest.log](https://github.com/rhinstaller/kickstart-tests/files/10003053/kstest.log)
[virt-install.log](https://github.com/rhinstaller/kickstart-tests/files/10003054/virt-install.log)

matejak:
#### Description:

Archives and ready-to-use content use paths differently.

Archives get unpacked into a directory, where they need to be unpacked, analyzed, and cross-checked with e.g. the supplied content path, whereas ready-to-use content can be used directly.

As the current codebase doesn't untangle all possible ways how to obtain existing content in a way of decomposing those into layers, this change just makes the current code working at the expense of making it worse to maintain.

#### Rationale:

- Before this PR, the simple use case of hardening a system with the system content was not working.

- Fixes https://github.com/rhinstaller/kickstart-tests/issues/827

#### Review Hints:

The PR makes use of the `preinst_content_path` that refers to absolute filepaths. `content_path` was supplied by users in case of archives, and represented the relative path to the content file in the context of the archive. In order to make it work, conversion of absolute filenames to relative filenames in the dictionary holding file labels has been dropped, and a subsequent "absolutization" of paths could be dropped too.
However, as the code assumes that the contents of `content_path` is always relative in case of archives and RPMs, this had to be moved to the method `use_downloaded_content`.

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -25,6 +25,14 @@ def is_network(scheme):
         for net_prefix in data_fetch.NET_URL_PREFIXES)
 
 
+def path_is_present_among_paths(path, paths):
+    absolute_path = os.path.abspath(path)
+    for second_path in paths:
+        if absolute_path == os.path.abspath(second_path):
+            return True
+    return False
+
+
 class ContentBringer:
     CONTENT_DOWNLOAD_LOCATION = pathlib.Path(common.INSTALLATION_CONTENT_DIR)
     DEFAULT_SSG_DATA_STREAM_PATH = f"{common.SSG_DIR}/{common.SSG_CONTENT}"
@@ -170,7 +178,7 @@ class ContentBringer:
             raise content_handling.ContentCheckError(msg)
 
     def allow_one_expected_tailoring_or_no_tailoring(self, labelled_files):
-        expected_tailoring = self._addon_data.tailoring_path
+        expected_tailoring = self._addon_data.preinst_tailoring_path
         tailoring_label = CONTENT_TYPES["TAILORING"]
         if expected_tailoring:
             labelled_files = self.reduce_files(labelled_files, expected_tailoring, [tailoring_label])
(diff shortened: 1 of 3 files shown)


jan-cerny:
I was testing this by creating an update image from the branch of this PR and using it on the top of the RHEL 8.7 installation. I was using kickstart installation.

- content-type = scap-security-guide works ok in both text and graphical kickstart installation
- content-type = rpm works OK in the text mode
- but in graphical mode - content-type = rpm  doesn't work, the OAA section shows "Error loading content" and if I open the OAA section I can see there error message in the red background at the bottom that the profile anssi_bp28_minimal isn't defined in the content . Unfortunately, I don't see any useful error message in the /tmp/anaconda.log of the installing system.

There is the kickstart below that I use:


```
lang en_US.UTF-9
timezone Europe/Prague
rootpw aaaaa
bootloader --location=mbr
clearpart --initlabel --all
autopart --type=plain
skipx

url --url=http://INSERT_URL_HERE/released/rhel-8/RHEL-8/8.7.0/BaseOS/x86_64/os/
repo --name appstream --baseurl=http://INSERT_URL_HERE/released/rhel-8/RHEL-8/8.7.0/AppStream/x86_64/os/

%packages
vim
%end

%addon org_fedora_oscap
  content-type = rpm
  content-url = "http://192.168.122.1:8000/scap-security-guide-0.1.63-4.el8.noarch.rpm"
  content-path = usr/share/xml/scap/ssg/content/ssg-rhel8-ds.xml
  profile = anssi_bp28_minimal
%end
```

The text variant that worked fine differed only by adding `text` on line before `url`.

jan-cerny:
@matejak  has discovered that the problem is that the GUI mode doesn't support shortened profile IDs in the profile key in the org_fedora_ kickstart section. This isn't a problem of this PR, it seems to exist before, and I have reported it as issue in https://github.com/OpenSCAP/oscap-anaconda-addon/issues/226

jan-cerny:
so when using a full profile ID in the profile key in the kickstart it works, and see the comment above for other scenarios that I tried

jan-cerny (approved):

PR #227, opened 2022-11-22
Rhel9 content paths

matejak:
#### Description:

Port of #225 to RHEL9. From its description: 

Archives and ready-to-use content use paths differently.

Archives get unpacked into a directory, where they need to be unpacked, analyzed, and cross-checked with e.g. the supplied content path, whereas ready-to-use content can be used directly.

As the current codebase doesn't untangle all possible ways how to obtain existing content in a way of decomposing those into layers, this change just makes the current code working at the expense of making it worse to maintain.

#### Rationale:

- Before this PR, the simple use case of hardening a system with the system content was not working.

#### Review Hints:

The PR makes use of the `preinst_content_path` that refers to absolute filepaths. `content_path` was supplied by users in case of archives, and represented the relative path to the content file in the context of the archive. In order to make it work, conversion of absolute filenames to relative filenames in the dictionary holding file labels has been dropped, and a subsequent "absolutization" of paths could be dropped too.
However, as the code assumes that the contents of `content_path` is always relative in case of archives and RPMs, this had to be moved to the method `use_downloaded_content`.

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -239,7 +239,7 @@ class ContentBringer:
                 f"Expected a file {expected_path} to be part of the supplied content, "
                 f"but it was not the case, got only {list(labelled_files.keys())}"
             )
-            raise RuntimeError(msg)
+            raise content_handling.ContentHandlingError(msg)
         for path, label in labelled_files.items():
             if label in categories and not paths_are_equivalent(path, expected_path):
                 continue
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -46,6 +46,14 @@ def clear_all(data):
     data.dry_run = False
 
 
+def path_is_present_among_paths(path, paths):
+    absolute_path = os.path.abspath(path)
+    for second_path in paths:
+        if absolute_path == os.path.abspath(second_path):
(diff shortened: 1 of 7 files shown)


jan-cerny:
I have verified that this PR fixes the installation fail when using `content-type = scap-security-guide` in the `%addon org_fedora_oscap` section in the kickstart. I tested that with both text and GUI installation of RHEL 9.1. In both cases it worked as expected and the installations finished successfully. I used the following command: `virt-install --connect qemu:///system --name oaa_test_rhel9 --memory 4096 --vcpus 2 --disk size=8 --os-variant rhel9.1 --location http://YOUR_URL/released/rhel-9/RHEL-9/9.1.0/BaseOS/x86_64/os/ -x inst.ks=http://192.168.122.1:8000/rhel9_ks_with_ssg.cfg -x inst.updates=http://192.168.122.1:8000/update.img --network default`


Then, I checked the installation with `content-type = rpm` in the `%addon org_fedora_oscap` section in the kickstart.  I tested that with both text and GUI installation of RHEL 9.1. I used the scap-security-guide-0.1.64-1.fc36.noarch.rpm RPM as external content served via http. I used the same command as above, only with a different kickstart file name. These also went fine and worked as expected.

jan-cerny (approved):

PR #232, opened 2023-01-18
Fix url parsing bug in OAA

matejak:
#### Description:

Handle properly the exception that occurs when entering a content URL that doesn't contain `://` as described in https://bugzilla.redhat.com/show_bug.cgi?id=2148509

#### Rationale:

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2148509

#### Review Hints:

- Do a GUI install with kickstart that uses scap-security-guide content, but wrong profile. Then, the whole content selection dialog will behave in a weird way. If you don't supply a bad profile, then the dialog will work, and the BZ will be reproducible.
- With this patch, both issues are resolved.

```
%addon org_fedora_oscap
    content-type = scap-security-guide
    profile = anssi_bp28_minimall
%end
```

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -67,9 +67,14 @@ class ContentBringer:
 
     @content_uri.setter
     def content_uri(self, uri):
-        scheme, path = uri.split("://", 1)
-        self.content_uri_path = path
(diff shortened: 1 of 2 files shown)


jan-cerny (approved):
I confirm that the linked BZ isn't reproducible with this patch on RHEL 8.7.

PR #235, opened 2023-01-20
Add libreport-plugin-rhtsupport to essential packages

matejak:
#### Description:

A content profile may require package removal, while the GUI environment requires the package via the `initial-setup` dependency. This PR makes sure that users get a warning at setup-time rather than in the middle of an installation.

#### Rationale:

- The RHEL8 OSPP profile requires the package.
- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2150355

#### Review hints:

- Run a graphical installation (no special kickstart is required), select the stock RHEL8 8.7 content, and select a "Server with GUI" environment.
- Before this fix, there was no error shown when the package was requested to be removed, after the fix, the situation is the same as with e.g. `nfs-tools`.

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -42,7 +42,9 @@ __all__ = ["RuleData"]
 
 # Mapping of packages to package environments and/or groups that depends on them
 # See also https://access.redhat.com/solutions/1201413 how to get group IDs.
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
I have verified that an error message about this offending package `libreport-plugin-rhetsupport` now appears in the `Changes that were done or need to be done` pane of the Security policy spokes when I chose to install OSPP with Server with GUI when I install RHEL 8.7 virtual machine.

PR #237, opened 2023-03-06
Refactor content fetching

jan-cerny:
#### Description:
This PR contains a series of small refactoring of the code responsible for content fetching, with focus on the `org_fedora_oscap.content_discovery.ContentBringer` class.

The main change is that the code performing analysis of the content has been moved away from `ContentBringer` to a new class `ContentAnalyzer`.

Moreover, code manipulating the members of the `PolicyData` class has been moved away from `ContentBringer` to the `PolicyData` class.

The outcome is that now the `ContentBringer` contains only code performing the download of the content and `ContentAnalyzer` contains only code performing the analysis of the already fetched content. These 2 classes are independent.

For more details, please read commit messages of all commits.

#### Rationale:
This change prepares the code for further refactoring with the goal to consolidate content fetching functionality. The separation of content fetching and content analysis functionality will allow us to focus on each of these 2 stages separately, with lower mutual collisions, in future PRs.

#### Review Hints:

cd testing_files/
python3 -m http.server

./create_update_image.sh install_addon_from_repo

./reanaconda.py updates ~/work/git/oscap-anaconda-addon/update.img

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -162,7 +162,7 @@ class ContentBringer:
             self._finish_actual_fetch(fetching_thread_name)
             if fingerprint and dest_filename:
                 self._verify_fingerprint(fingerprint)
-            content = self._analyze_fetched_content(
+            content = ContentBringer.__analyze_fetched_content(
                 fetching_thread_name, fingerprint, dest_filename,
                 expected_path, expected_tailoring, expected_cpe_path)
         except Exception as exc:
@@ -239,13 +239,14 @@ class ContentBringer:
             reduced_files[path] = label
         return reduced_files
 
-    def _analyze_fetched_content(
-                self, wait_for, fingerprint, dest_filename, expected_path,
+    @staticmethod
+    def __analyze_fetched_content(
+                wait_for, fingerprint, dest_filename, expected_path,
                 expected_tailoring, expected_cpe_path):
         actually_fetched_content = wait_for is not None
         fpaths = ContentBringer.__gather_available_files(actually_fetched_content, dest_filename)
 
-        structured_content = ObtainedContent(self.CONTENT_DOWNLOAD_LOCATION)
+        structured_content = ObtainedContent(ContentBringer.CONTENT_DOWNLOAD_LOCATION)
         content_type = ContentBringer.__get_content_type(str(dest_filename))
         log.info(f"OSCAP Addon: started to look at the content")
         if content_type in ("archive", "rpm"):
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -290,7 +290,7 @@ class ContentBringer:
         return fpaths
 
     def use_downloaded_content(self, content):
-        preferred_content = self.get_preferred_content(content)
+        preferred_content = self.get_preferred_content(self._addon_data.content_path, content)
 
         # We know that we have ended up with a datastream-like content,
         # but if we can't convert an archive to a datastream.
@@ -309,9 +309,9 @@ class ContentBringer:
             else:
                 self._addon_data.tailoring_path = str(preferred_tailoring)
 
-    def get_preferred_content(self, content):
-        if self._addon_data.content_path:
-            preferred_content = content.find_expected_usable_content(self._addon_data.content_path)
+    def get_preferred_content(self, content_path, content):
+        if content_path:
+            preferred_content = content.find_expected_usable_content(content_path)
         else:
             preferred_content = content.select_main_usable_content()
         return preferred_content
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -217,10 +217,10 @@ class ContentBringer:
         expected_tailoring = common.get_preinst_tailoring_path(self._addon_data)
         labelled_files = self.allow_one_expected_tailoring_or_no_tailoring(labelled_files, expected_tailoring)
 
-        expected_path = self._addon_data.cpe_path
+        expected_cpe_path = self._addon_data.cpe_path
         categories = (CONTENT_TYPES["CPE_DICT"], )
-        if expected_path:
-            labelled_files = self.reduce_files(labelled_files, expected_path, categories)
+        if expected_cpe_path:
+            labelled_files = self.reduce_files(labelled_files, expected_cpe_path, categories)
 
         return labelled_files
 
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -159,7 +159,7 @@ class ContentBringer:
             )
         return fetching_thread_name
 
-    def finish_content_fetch(self, fetching_thread_name, fingerprint, report_callback, dest_filename,
+    def finish_content_fetch(self, fetching_thread_name, fingerprint, dest_filename,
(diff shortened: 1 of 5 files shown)


org_fedora_oscap/content_discovery.py
```diff
@@ -136,10 +136,9 @@ def _fetch_files(self, scheme, path, destdir, ca_certs_path, what_if_fail):
         # We are not finished yet with the fetch
         return fetching_thread_name
 
-    def _start_actual_fetch(self, scheme, path, destdir, ca_certs_path):
-        fetching_thread_name = None
-        url = scheme + "://" + path
-
+    @staticmethod
+    def __get_dest_file_name(url, destdir):
+        path = url.split("://")[1]
```
matejak: This will crash in some cases of invalid URLs s.a. `http://`. We had a BZ about this recently, and the code that extracts path in a robust way is already in the project, that could be extracted as well.
The code also assumes that destdir is a pathlib path, so adding an appropriate type hint may be a good idea.

org_fedora_oscap/content_discovery.py
```diff
@@ -84,7 +83,7 @@ def get_content_type(self, url):
 
     @property
     def content_uri(self):
-        return self.content_uri_scheme + "://" + self.content_uri_path
+        return self._content_uri
```
matejak: I suggest renaming to `_valid_content_uri`, so the code that uses it knows that it is safe to make certain assumptions.

org_fedora_oscap/content_discovery.py
```diff
@@ -93,7 +94,16 @@ def content_uri(self, uri):
...
+        path = scheme_and_maybe_path[1]
+        if "/" not in path:
+            msg = f"Missing the path component of the '{uri}' URL"
+            raise KickstartValueError(msg)
+        basename = path.rsplit("/", 1)[1]
+        if not basename:
+            msg = f"Unable to deduce basename from the '{uri}' URL"
+            raise KickstartValueError(msg)
         self._content_uri = uri
+        self.dest_file_name = self.CONTENT_DOWNLOAD_LOCATION / basename
```
matejak: Why not to keep the code in that private method, and call it only once in the constructor?

org_fedora_oscap/content_discovery.py
```diff
@@ -159,7 +159,7 @@ def _start_actual_fetch(self, ca_certs_path):
             )
         return fetching_thread_name
 
-    def finish_content_fetch(self, fetching_thread_name, fingerprint, report_callback, dest_filename,
+    def finish_content_fetch(self, fetching_thread_name, fingerprint, dest_filename,
```
matejak: This is tricky, the report callback is supposed to be a function that I can use to get information out of the depths of data processing to the user in a good form. It may not be used because of simple omission, not because it doesn't make sense to use it. IOW, the fix for this unused argument could be that it is put to a use.

org_fedora_oscap/structures.py
```diff
@@ -202,11 +200,5 @@ def clear_all(self):
...
         self.cpe_path = ""
         self.tailoring_path = ""
-
         self.fingerprint = ""
-
         self.certificates = ""
-
-        # internal values
-        self.rule_data = rule_handling.RuleData()
-        self.dry_run = False
```
matejak: This is interesting, almost suspicious. Note to self - needs further investigation.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -1215,6 +1215,6 @@ def on_change_content_clicked(self, *args):
         self._refresh_ui()
 
     def on_use_ssg_clicked(self, *args):
-        self.content_bringer.use_system_content()
```
matejak, line 1217: I disagree here - the content spec data class shouldn't have the knowledge what "system content" even is, and even how to get it. The "bringer" class knows this, and, of course, hammers the information into the content spec aspect of the "policy data" dataclass. It is OK for the bringer to know the data format of the content spec, and the content spec just has to store the data, it doesn't have to fully understand them to the finest extent.

org_fedora_oscap/structures.py
```diff
@@ -209,3 +209,22 @@ def use_system_content(self):
         self.clear_all()
         self.content_type = "scap-security-guide"
         self.content_path = common.get_ssg_path()
+
+    def use_downloaded_content(self, content):
```
matejak: Ditto, the logic of obtaining content is independent from the description of the intent. The PolicyData should store data that have objective meaning, and a single line of its definition shouldn't need to be touched if we decide to use different algorithm for content download, or other implementation detail.

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -436,14 +436,14 @@ def update_progress_label(msg):
 
         if actually_fetched_content:
             content_path = common.get_raw_preinst_content_path(self._policy_data)
+            self.content_bringer.finish_content_fetch(
```
matejak, line 436: Note for self: Probably the right thing to do now, but we need to change how it works.
Code should read like a prose, not like a cipher game, as it reads right now.

org_fedora_oscap/service/installation.py
```diff
@@ -96,9 +96,10 @@ def run(self):
         expected_path = common.get_preinst_content_path(self._policy_data)
         expected_tailoring = common.get_preinst_tailoring_path(self._policy_data)
         expected_cpe_path = self._policy_data.cpe_path
-        self.content_bringer.finish_content_fetch(
-            fetching_thread_name, self._policy_data.fingerprint,
-            content_dest, _handle_error)
+        if fetching_thread_name is not None:
```
matejak, line 101: Note to self: Probably the right thing to do now, but we need to change how it works.
Code should read like a prose, not like a cipher game, as it reads right now. It is very well visible here that a variable storing a name of a thread is seriously abused.

matejak (changes requested):
Thank you for the PR, it does many changes totally right, while some I don't like, so let's talk about them.

Great:

- Streamlining of the code.
- Factoring the analyzer out of the bringer.

Controversial:

- PolicyData getting new abilities. I agree with the ability to "clear" itself, but the rest should be responsibility of a different class. PolicyData should remain a data structure, an interface that other classes understand and use. This data structure shouldn't know that a thing s.a. "system content" or "downloadable content" exist.

org_fedora_oscap/content_discovery.py
jan-cerny: Good point, but in later commits in this PR, the ` __get_dest_file_name` is further removed and the code using the URI is consolidated at a single place: the `content_uri` setter method, in which the sanity of the URI is checked and an exception is raised.

org_fedora_oscap/content_discovery.py
jan-cerny: Because the basename is derived from URI of the content and the URI will change each time a new content will be requested to fetch.

org_fedora_oscap/content_discovery.py
jan-cerny: Do you have an idea what information it should return?

jan-cerny:
I have moved the methods out from the `PolicyData` class and then I have renamed `_content_uri`.

jan-cerny:
@matejak ^

matejak (approved):

matejak:
LGTM, thanks!

PR #239, opened 2023-03-27
Introduce policy data wrapper class

jan-cerny:
#### Description:

Introduce a PolicyDataHandler class and move some items from the OSCAPScope class to this new class.

#### Rationale:
OSCAPScope is a class that handles GUI, but has too many other responsibilities, apart from GUI items there is many functional logic involved. The OSCAPScope often manipulates the PolicyData class, so it would be natural to move this code to the PolicyData class. However, this class is intended to work only as a data structure. We will not break this intention so we will create a new class PolicyDataHandler that manipulates the existing PolicyData class.

This separation also improves testability.

#### Review Hints:
all relates to GUI installation
./create_update_image.sh install_addon_from_repo
./reanaconda.py updates ~/work/git/oscap-anaconda-addon/update.img 
try to load content from a server, eg. provided by `python3 -m http.server`

### DIFF
--- a/org_fedora_oscap/gui/spokes/oscap.py
+++ b/org_fedora_oscap/gui/spokes/oscap.py
@@ -33,6 +33,7 @@ from org_fedora_oscap import scap_content_handler
 from org_fedora_oscap import content_discovery
 from org_fedora_oscap import utils
 from org_fedora_oscap.constants import OSCAP
+from org_fedora_oscap.data_handling import DataHandler
 from org_fedora_oscap.structures import PolicyData
 
 from pyanaconda.modules.common.constants.services import USERS
@@ -240,6 +241,7 @@ class OSCAPSpoke(NormalSpoke):
 
         self._policy_data = PolicyData()
         self._load_policy_data()
+        self.data_handler = DataHandler(self._policy_data)
 
         # used for changing profiles
         self._rule_data = None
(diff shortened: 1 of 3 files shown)


marcusburghardt:
@matejak , could you help reviewing this PR, please? I believe you are more familiar with OAA. Thanks

org_fedora_oscap/gui/spokes/oscap.py
```diff
@@ -402,7 +399,7 @@ def _fetch_data_and_initialize(self):
             self._fetching = True
 
         thread_name = None
-        if self._policy_data.content_url and self._policy_data.content_type != "scap-security-guide":
+        if self.data_handler.needs_fetch_content():
```
matejak: I suggest to rename `needs_fetch_content` to `content_is_fetchable` or something like that, because the need arises from the fact that we don't have it, which may not be the case.

matejak:
This is great, it makes the code much more readable, and extracts reusable parts, so they can be reused, at least in theory.
Could you please check out whether this approach couldn't benefit functionality in https://github.com/OpenSCAP/oscap-anaconda-addon/blob/rhel9-branch/org_fedora_oscap/service/installation.py ? I think that the approach of fetching in https://github.com/OpenSCAP/oscap-anaconda-addon/blob/rhel9-branch/org_fedora_oscap/service/installation.py#L86 is wrong, and `needs_fetch_content` functionality could be useful here, because if we use system content, there is no need to even try to download anything. I think that now, some sort of fetching is attempted even in the case of the system content, and inevitable errors are just silently ignored.

jan-cerny:
I have renamed the aforementioned method and then I have changed the condition in PrepareValidContent.run().

jan-cerny:
@matejak what else needs to be done?

matejak (approved):
Great, thanks!

PR #240, opened 2023-04-14
Make the Bringer better testable

matejak:
#### Description:

This PR moves functionality formerly located in the data_fetch module to the Bringer, so one can use polymorphism to create great mocks by just subclassing.
Now, it is possible to test handling of threads and fetch methods separately, because Bringer just calls a method to fetch a URI, and then wraps the process in a thread.

#### Rationale:

- The PR decomposes and concentrates the existing functionality, making it easier to test.

#### Review Hints:

- Business as usual, tests pass, bringer still able to copy things, touched functions were not used by other parts of the addon.

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -43,6 +43,7 @@ class ContentBringer:
 
     def __init__(self, what_if_fail):
         self._valid_content_uri = ""
+        self.ca_certs_path = ""
         self.dest_file_name = ""
 
         self.activity_lock = threading.Lock()
@@ -84,14 +85,15 @@ class ContentBringer:
         """
         try:
             self.content_uri = content_uri
+            self.ca_certs_path = ca_certs_path
         except Exception as exc:
             self.what_if_fail(exc)
         shutil.rmtree(self.CONTENT_DOWNLOAD_LOCATION, ignore_errors=True)
         self.CONTENT_DOWNLOAD_LOCATION.mkdir(parents=True, exist_ok=True)
-        fetching_thread_name = self._fetch_files(ca_certs_path)
+        fetching_thread_name = self._fetch_files()
         return fetching_thread_name
 
-    def _fetch_files(self, ca_certs_path):
+    def _fetch_files(self):
         with self.activity_lock:
             if self.now_fetching_or_processing:
                 msg = _(
@@ -103,7 +105,7 @@ class ContentBringer:
 
         fetching_thread_name = None
         try:
-            fetching_thread_name = self._start_actual_fetch(ca_certs_path)
+            fetching_thread_name = self._start_actual_fetch()
         except Exception as exc:
             with self.activity_lock:
                 self.now_fetching_or_processing = False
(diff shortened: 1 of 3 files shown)


org_fedora_oscap/content_discovery.py
```diff
@@ -112,23 +113,31 @@ def _fetch_files(self, ca_certs_path):
...
+            try:
+                data_fetch.wait_for_network()
+            except common.OSCAPaddonNetworkError:
+                msg = _("Network connection needed to fetch data.")
+                raise common.OSCAPaddonNetworkError(msg)
+
+        fetch_data_thread = AnacondaThread(
+            name=fetching_thread_name,
+            target=self.fetch_operation,
+            args=(self.content_uri, self.dest_file_name, ca_certs_path),
```
jan-cerny: I have noticed a weird thing that was there before but I think we can talk about it nevertheless. The URI is a property of the class (`self.content_uri`) but the certificate path isn't a property of the class but a variable (`ca_certs_path`) and is passed down. But, both are arguments of the `fetch_content` content method. So the arguments of `fetch_content` are each handled differently at this moment.

org_fedora_oscap/content_discovery.py
```diff
@@ -112,23 +113,31 @@ def _fetch_files(self, ca_certs_path):
...
+            fatal=False)
+
+        # register and run the thread
+        threadMgr.add(fetch_data_thread)
+
         return fetching_thread_name
 
-    def finish_content_fetch(self, fetching_thread_name, fingerprint):
+    def fetch_operation(self, uri, out_file, ca_certs_path=None):
+        return data_fetch.fetch_data(uri, out_file, ca_certs_path)
```
jan-cerny: What is the reason for having this wrapper function? Can we simply move code from the `data_fetch.fetch_data` to `fetch_operation`?

jan-cerny (commented):
So far it works great for me because I tried to build the update image and run the Anaconda installation and changed various SCAP content in the Security Policy Scope.

However, I have found an issue, which isn't caused by the contents of this PR: https://github.com/OpenSCAP/oscap-anaconda-addon/issues/244

org_fedora_oscap/content_discovery.py
matejak: Good point, the common argument to a generl fetch operation would be the URI that we store in the class anyway, and then other arguments depend on how we fetch, certs being needed only if we fetch from network using an "s" protocol.
Therefore, it is cleaner to store those variable arguments in the instance as well.

org_fedora_oscap/content_discovery.py
matejak: The other function is somewhat coupled with the `data_fetch` module - the way it is unit-tested as well as the way exceptions it can raise are defined.
The whole data_fetch module could be merged into the bringer in the future, but I think that it is OK to have such thin wrapper.
The fetch operation as a separate method is convenient to be able to override it in child classes e.g. for testing purposes.

jan-cerny (approved):
I have tried to download various content usign the GUI and I have performed a RHEL9.1 installation from GUI and also an unattended installation with a kickstart.

PR #242, opened 2023-04-20
Port #237 to rawhide branch

jan-cerny:
This ports https://github.com/OpenSCAP/oscap-anaconda-addon/pull/237 to `rawhide` branch.

I'm unable to say if it works or not. How can I test this? The steps in the Developer guide suggests using `reanaconda`, but `reanaconda prime` fails with a traceback when I try to pass a Fedora rawhide URL installation tree.

### DIFF
--- a/org_fedora_oscap/content_discovery.py
+++ b/org_fedora_oscap/content_discovery.py
@@ -93,11 +93,10 @@ class ContentBringer:
         self.content_uri = self._addon_data.content_url
         shutil.rmtree(self.CONTENT_DOWNLOAD_LOCATION, ignore_errors=True)
         self.CONTENT_DOWNLOAD_LOCATION.mkdir(parents=True, exist_ok=True)
-        fetching_thread_name = self._fetch_files(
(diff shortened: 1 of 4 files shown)


matejak:
First of all, the test fails for reasons that should be fixable in the PR.

Regarding integration testing, a downloaded Fedora image + specifying `inst.updates=http://...` in the VM works, although it is much more cumbersome. However, the Fedora `os` tree is something else than the RHEL tree, so it is hard to say whether it can even be fixed.

jan-cerny:
@jan-cerny Thanks for suggestion. I believe we should update the developer documentation then.

jan-cerny:
I have fixed the failed unit test, I have rebased this PR on the top of the `rawhide` branch and I have also successfully executed it in the Fedora install on the Fedora Server 38 installation DVD iso.

jan-cerny:
The CI fail is unrelated to the contents of this PR because we can see the same CI fail in #245 , it looks like the job is broken in general. Any thoughts?

jan-cerny:
The CI fail seems related to #238.

matejak:
Please try to rebase, tests should pass now.

PR #248, opened 2023-07-17
Remove obsolete mapping of packages-groups

matejak:
#### Description:

The mapping is a pain point in the addon - the related code is basically a cache of information that primarily exists in package/group metadata.

#### Rationale:

The `sftp` package is not part of the group any more, and `abrt` is not part of anything, because it doesn't exist.

- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2172264

#### Review Hints:

Run e.g. `dnf group info network-server | grep ftp` on a RHEL9 system, or explore the package-group mapping by other means.

### DIFF
--- a/org_fedora_oscap/rule_handling.py
+++ b/org_fedora_oscap/rule_handling.py
@@ -59,12 +59,6 @@ ESSENTIAL_PACKAGES = {
         "env": ["graphical-server-environment", "workstation-product-environment"],
         "groups": ["workstation-product-environment"],
     },
-    "tftp": {
-        "groups": ["network-server"],
-    },
-    "abrt": {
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
I have started a RHEL 9.2 installation using the `reanaconda` script. I have reproduced https://bugzilla.redhat.com/show_bug.cgi?id=2172264. I confirm that this PR has fixed this BZ for me. The message "package 'tftp' has been added to the list of excluded packages, but it can't be removed from the current software selection without breaking the install" no longer appears with this fix, and instead "package 'tftp' has been added to the list of excluded packages" appeared. Great job!

PR #249, opened 2023-07-17
Make tar extraction safer on RHEL8

matejak:
See also https://bugzilla.redhat.com/show_bug.cgi?id=2219408

According to the referenced BZ, RHEL 8.9+ should contain means that allow safer tarfile extraction

#### Review Hints:

Test an install that uses a tar file as content container

### DIFF
--- a/org_fedora_oscap/common.py
+++ b/org_fedora_oscap/common.py
@@ -360,7 +360,7 @@ def extract_data(archive, out_dir, ensure_has_files=None):
                 raise ExtractionError(msg)
 
         utils.ensure_dir_exists(out_dir)
-        zfile.extractall(path=out_dir)
+        zfile.extractall(path=out_dir, filter="data")
         result = [utils.join_paths(out_dir, info.filename) for info in zfile.filelist]
         zfile.close()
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
I've been able to install a RHEL 9.3 successfully with a content with a tarball, using the `reanaconda` tool, using the following KS:


```
lang en_US.UTF-9
timezone Europe/Prague
rootpw aaaaa
bootloader --location=mbr
clearpart --initlabel --all
autopart --type=plain
text

url --url=http://XXXXX/rhel-8/nightly/RHEL-8/latest-RHEL-8.9/compose/BaseOS/x86_64/os/
repo --name appstream --baseurl=http://XXXXX/rhel-8/nightly/RHEL-8/latest-RHEL-8.9/compose/AppStream/x86_64/os/

%packages
vim
%end

%addon org_fedora_oscap
    content-url = http://10.0.2.2:8000/ssgrhel8.tar.gz
    content-type = archive
    content-path = ssg-rhel8-ds.xml
    profile = xccdf_org.ssgproject.content_profile_anssi_bp28_minimal
%end
```

PR #255, opened 2024-10-04
OPENSCAP-4522 Do not assume availability of hashing algorithms in hashlib

evgenyz:
Particular offender at this moment is 'md5', which is not available in FIPS build of Python.

See: https://issues.redhat.com/browse/RHEL-40367.


Mab879 (approved):
Thanks!

